PR #11, opened 2016-01-28
Removed deprecated and useless classes.

carturn:
The RobotBase class is confusing.
LIDAR should be handled off-rio.
I do not think anyone uses the IMU class and it has very 2015 specific code.


carturn:
This has been tested.

c0d3rman:
+1

PR #12, opened 2016-01-29
Can leds

carturn:
Added CustomNeoPixels.
This allows us to control LEDs over CAN.


carturn:
For the record, I tested this yesterday. It works and is super cool.

PR #24, opened 2016-02-05
Fixed call hierarchy in logkitten.

carturn:
The level-log methods without overrides had a different call hierarchy level, which broke the stack-trace usage.


ajnadel:
+1 once tested.

PR #28, opened 2016-02-10
Added more base Commands

c0d3rman:
Added more base commands:

- Cancel cancels a given command on initialize
- Idle does nothing but requires the given subsystem (intended to be used as a default command, which must require the subsystem)
- RunFor runs the given command for a given duration and then cancels it
- RunIf uses anonymous interfaces (optionally made a LOT cleaner by the Java 8 double colon; see comments in code) to run the given command only if the given function(s) returns true.
- RunUnless does the opposite of RunIf


carturn:
addSequential needs to be in the constructor

carturn:
Also, I maintain that we should try to avoid Java 8 syntax.

c0d3rman:
The documentation on the RunIf and RunUnless commands gives a Java 7- alternative, but the Java 8 syntax is cleaner and better.
Also, does it have to be? If I move it to the constructor it will evaluate the check at construct-time, not initialize-time, so I'd have to skip the CommandGroup structure and manually spawn the command (which would have other downsides, like not using CommandGroup's requirement handling.)

carturn:
The code will literally not run unless addSequential is in the constructor. The command will not initialize until the CommandGroup initializes though.

carturn:
I like all of these except RunIf and RunUnless. I also feel like RunFor should be done by extending whatever class you are running for some time.

c0d3rman:
OK, this is problematic - we can't do the check at initialize time if we addSequential at construct-time.
However, RunIf is going to be important for CommandGroups like Shoot that need to call things like flywheel.isSpunUp and hood.isUp at initialize time. I don't want a RockToShootIfSpunUpAndHoodUp command.

carturn:
I think RunIf and RunUnless are just going to have to be more complicated commands.

carturn:
Shoot could be a commandGroup with sequential hood up, then spin up, then rock to shoot.

carturn:
Also, the existence of a BooleanInterface bothers me.

c0d3rman:
But that's not what the trigger does. In our control scheme the trigger rocks to the shoot position iff the hood is up and the flywheel is spun up. So Shoot cannot do those things, it can only check for them.

carturn:
Fine, so Shoot has to take in the hood and the flywheel and do checks on them.

carturn:
I think that is probably the best way to do that. I agree that it would be better to have a generic version, but I think it is more complicated than it is worth.

carturn:
Also, when you end a commandGroup, it automatically cancels all subcommands.

c0d3rman:
Shoot already takes them in, but cannot run checks on them - that is the problem I was addressing when I created these commands. Shoot is a CommandGroup and so must `addSequential` at construct-time. It can't check the flywheel and hood at runtime.

carturn:
Why is shoot a command group?

carturn:
Should not it just be RockToShoot?

c0d3rman:
It has to check whether it can shoot (and possibly send some signals to the DS for the color-coding).

carturn:
:+1:

c0d3rman:
:+1:

carturn:
Idle, Cancel, and RunFor works.

carturn:
RunIf, RunUnless, and RunIfElse work.

carturn:
:+1:

c0d3rman:
:+1:

PR #29, opened 2016-02-11
Superbutton enhancement

c0d3rman:
Improved SuperButton.


carturn:
:+1: once tested.

carturn:
I maintain :+1: once tested. I added some stuff to use CustomButtons in CustomXbox. Also, they should be called CustomButtons, not SuperButtons, to avoid confusion with java super keyword. We have been using Custom as the standard elsewhere in standard.

carturn:
cancelWhenReleased and onlyWhileHeld work, but onlyWhileReleased starts running after you release the button and stops running when you press it again. I am not sure if this is the intended behavior.

c0d3rman:
onlyWhileReleased should have the exact opposite effect of
onlyWhilePressed. So what you are describing is correct, except it should
run the command before the first press of the button.

carturn:
@c0d3rman I do not think that is possible because the command (or CustomButton) would have to detect when the robot is enabled and start the command then.

carturn:
:+1:

carturn:
"Improved Javadocing" also made the commands start automatically.

carturn:
:+1:

c0d3rman:
:+1:

PR #30, opened 2016-02-11
Removed ParticleReport.

carturn:


c0d3rman:
:+1:. I think we can bypass testing, since the class is never used.

PR #31, opened 2016-02-12
Sensor motors

carturn:
I think we are ready to test and merge this. It will be needed for good shooting.


carturn:
CAN sensors is now part of this.

c0d3rman:
:+1:

PR #37, opened 2016-02-14
Pid chassis controller

tobbrau:
To be tested. This uses pid as a chassis controller, it has been generalized to use any NavX


tobbrau:
NEEDS TO BE TESTED

c0d3rman:
:+1: even though we never fixed the high gear constants

PR #41, opened 2016-02-17
CANTalon encoder

carturn:
Basically just reading the encoder over the talon.

### DIFF
--- a/custom/sensors/TalonEncoder.java
+++ b/custom/sensors/TalonEncoder.java
@@ -1,0 +1,89 @@
+package org.usfirst.frc4904.standard.custom.sensors;
+
+
+import edu.wpi.first.wpilibj.CANTalon;
+import edu.wpi.first.wpilibj.PIDSourceType;
+
+public class TalonEncoder implements CustomEncoder {
+	protected final CANTalon talon;
+	private PIDSourceType pidSource;
+	private double distancePerPulse;
+	private boolean reverseDirection;
+	private boolean rateMode;
+	
+	public TalonEncoder(String name, CANTalon talon, boolean reverseDirection, double distancePerPulse, boolean rateMode) {
+		this.talon = talon;
+		this.reverseDirection = reverseDirection;
+		this.distancePerPulse = distancePerPulse;
+		this.rateMode = rateMode;
+		setPIDSourceType(PIDSourceType.kDisplacement);
+	}
+	
+	@Override
+	public void setPIDSourceType(PIDSourceType pidSource) {
+		this.pidSource = pidSource;
+	}
+	
+	@Override
+	public PIDSourceType getPIDSourceType() {
+		return pidSource;
+	}
+	
+	@Override
+	public double pidGet() {
+		if (rateMode) {
+			return getRate();
+		}
+		return getDistance();
+	}
+	
+	@Override
+	public int get() {
+		return talon.getEncPosition();
+	}
+	
+	@Override
+	public double getDistance() {
+		if (reverseDirection) {
+			return distancePerPulse * (double) talon.getEncPosition() * -1.0;
+		} else {
+			return distancePerPulse * (double) talon.getEncPosition();
+		}
+	}
+	
+	@Override
+	public boolean getDirection() {
+		return !reverseDirection == (talon.getEncVelocity() >= 0);
+	}
+	
+	@Override
+	public boolean getStopped() {
+		return Math.abs(getRate()) <= 0.0001;
+	}
+	
+	@Override
+	public double getRate() {
+		if (reverseDirection) {
+			return talon.getEncVelocity() * -10.0 * distancePerPulse;
+		} else {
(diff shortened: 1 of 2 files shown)


custom/sensors/TalonEncoder.java
```diff
@@ -0,0 +1,89 @@
+package org.usfirst.frc4904.standard.custom.sensors;
+
+
+import edu.wpi.first.wpilibj.CANTalon;
+import edu.wpi.first.wpilibj.PIDSourceType;
+
+public class TalonEncoder implements CustomEncoder {
```
c0d3rman: This should be refactored to CANTalonEncoder to indicate that it only works with TalonSRX's (i.e. Talons you use over CAN).

custom/sensors/TalonEncoder.java
```diff
@@ -0,0 +1,89 @@
...
+	protected final CANTalon talon;
+	private PIDSourceType pidSource;
```
c0d3rman: Why are these private? Should they not be protected? I can't foresee anything extending this but the philosophy is don't make it private unless you specifically want to disallow access to extending classes.

custom/sensors/TalonEncoder.java
```diff
@@ -0,0 +1,89 @@
...
+		this.pidSource = pidSource;
+	}
+	
+	@Override
+	public PIDSourceType getPIDSourceType() {
+		return pidSource;
+	}
+	
+	@Override
+	public double pidGet() {
```
c0d3rman: I think we should piggyback off of PIDSourceType, PIDSource, and PIDOutput for our PID framework - the framework makes sense save for PIDController. CustomPID, calling pidGet, can also figure out whether its source is returning rate or displacement, and then we set the output type as well as the input type (this is what PIDController is missing.)

custom/sensors/TalonEncoder.java
```diff
@@ -0,0 +1,89 @@
...
+	}
+	
+	@Override
+	public boolean getDirection() {
+		return !reverseDirection == (talon.getEncVelocity() >= 0);
+	}
+	
+	@Override
+	public boolean getStopped() {
+		return Math.abs(getRate()) <= 0.0001;
```
c0d3rman: In the interest of limiting magic numbers, can we make this a constant?
Also, this needs to be way smaller and hopefully less arbitrary. Would Double.MIN_VALUE, Double.MIN_NORMAL, or Double.MIN_EXPONENT work?
Maybe we should have a system-wide epsilon.
Finally, we need to work on eliminating floating point errors from the other end. See Mark Jeronimus's answer [here](http://stackoverflow.com/questions/8194917/comparing-a-double-against-zero) - the jist is, do all of your calculations for the numerator and denominator and divide only once. It might make the calculations a little (read: a lot) uglier, though.

custom/sensors/TalonEncoder.java
```diff
@@ -0,0 +1,89 @@
...
+	}
+	
+	@Override
+	public double getRate() {
+		if (reverseDirection) {
+			return talon.getEncVelocity() * -10.0 * distancePerPulse;
```
c0d3rman: Why is this -10.0 and 10.0? Shouldn't it be -1.0 and 1.0?

custom/sensors/TalonEncoder.java
c0d3rman: This would replace rateMode here, but not on the PIDController (it would have to be refactored).

custom/sensors/TalonEncoder.java
carturn: I feel like CAN is implied, but maybe.

custom/sensors/TalonEncoder.java
carturn: getEncVelocity is in ticks per 100ms, getRate returns in ticks per second.

c0d3rman: Ah. Can we comment that in? I don't think it needs to go so far as a constant.

custom/sensors/TalonEncoder.java
carturn: Resolved.

custom/sensors/TalonEncoder.java
carturn: Resolved.

custom/sensors/TalonEncoder.java
c0d3rman: Another [SO post](http://stackoverflow.com/questions/356807/java-double-comparison-epsilon) on the topic. Maybe we should use an arbitrary precision number format for the PID loop internally, but probably overkill.

carturn: It would be nice if there was a Java standard value. Maybe we could create one, but I am not sure where to put it. It would be nice to have a standard generic things class though (I want a get delta time function).

c0d3rman:
:+1:

PR #42, opened 2016-02-17
Removed obsolete rateMode boolean.

carturn:
This removes rateMode boolean, people should use PIDSourceType instead.
CustomPID makes this reasonable.

### DIFF
--- a/custom/sensors/CANEncoder.java
+++ b/custom/sensors/CANEncoder.java
@@ -66,13 +66,10 @@ public class CANEncoder extends CANSensor implements CustomEncoder {
 	}
 	
 	public double pidGet() {
-		switch (pidSource) {
-			case kRate:
-				return getRate();
-			case kDisplacement:
-			default:
-				return getDistance();
+		if (pidSource == PIDSourceType.kDisplacement) {
+			return getDistance();
 		}
+		return getRate();
 	}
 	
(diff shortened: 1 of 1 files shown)


custom/sensors/CANEncoder.java
```diff
@@ -88,10 +66,13 @@ public void setReverseDirection(boolean reverseDirection) {
 	}
 	
 	public double pidGet() {
-		if (rateMode) {
-			return getRate();
+		switch (pidSource) {
```
c0d3rman: Is a switch statement really necessary? PIDSourceType is an enum.

custom/sensors/CANEncoder.java
```diff
@@ -12,58 +12,36 @@
 	private PIDSourceType pidSource;
```
c0d3rman: I don't like this name, but it's the one Encoder uses in wpilib, so OK.

custom/sensors/CANEncoder.java
carturn: YES

custom/sensors/CANEncoder.java
carturn: OK.

c0d3rman:
:+1:

PR #48, opened 2016-02-18
don't log the entire package in LogKitten

ajnadel:
This PR:
- doesn't log the class's entire package in LogKitten


c0d3rman:
@ajnadel Is this tested? If so, please add the "tested" label. In the meantime, :+1:.

ajnadel:
@c0d3rman

PR #50, opened 2016-02-18
Standard standards

carturn:
Adding a set of common miscellaneous capabilities.
Currently has a zero checking system.


Util.java
```diff
@@ -0,0 +1,8 @@
+package org.usfirst.frc4904.standard;
+
+
+public class Util {
+	public static boolean checkZero(double var) {
```
c0d3rman: Refactor to isZero?

c0d3rman: Or perhaps isZeroFPASafe or something else to indicate when/why it should be used. Definitely Javadoc it.

carturn: Resolved.

c0d3rman:
:+1:

PR #52, opened 2016-02-18
Generic motion controllers

carturn:
This change replaces CustomPID with MotionControllers, a generic motion controlling system.
This is designed to make using BangBang, PID, etc. relatively interchangable.


c0d3rman:
:+1:
=

carturn:
:+1:

PR #55, opened 2016-02-18
SafetyCommand constructors

ajnadel:
This PR:
- adds Command's constructors to SafetyCommand

I don't think it needs to be tested, due to the nature of the changes.


c0d3rman:
:+1:

PR #57, opened 2016-02-19
CANTalonEncoder remove rateMode

c0d3rman:
Removed rateMode from CANTalonEncoder and added convenience constructors.


carturn:
:+1:

PR #59, opened 2016-02-19
Speed modifier group

carturn:
This adds the Speed Modifier Group, allowing one to apply multiple speed modifiers to one motor.


c0d3rman:
:+1:

PR #70, opened 2016-03-05
Synchronized logkitten logging methods

c0d3rman:
This PR synchronizes all of the LogKitten logging methods.
This mirrors the changes on `skynet-sockets`.
@HowardStark please devise a way to test this and report back.

### DIFF
--- a/LogKitten.java
+++ b/LogKitten.java
@@ -92,7 +92,7 @@ public class LogKitten {
 	 * @param DEFAULT_LOG_LEVEL
 	 *        default write-to-file level
 	 */
-	public static void setDefaultLogLevel(KittenLevel DEFAULT_LOG_LEVEL) {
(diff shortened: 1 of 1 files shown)


c0d3rman:
Does not seem to break anything.

c0d3rman:
As @leijurv pointed out, this may cause issues because these methods are called many times a second (especially `d`) from different threads and write to files and hence this may create a backlog.
`logMessage` is notably not synchronized. @HowardStark – am I misunderstanding this?

leijurv:
All the methods that call logMessage don't need to be synchronized, it suffices to just have logMessage be synchronized. Methods that set fields to the argument don't need to be synchronized. It suffices to set a boolean as volatile, as that will automatically synchronize accesses.

leijurv:
+1

leijurv:
+1

leijurv:
+1

leijurv:
+1

PR #72, opened 2016-03-06
Fixed LogKitten exception override

carturn:
LogKitten exception logging now has override as a parameter, rather than overriding by default.


c0d3rman:
:+1:

c0d3rman:
Works as expected. (Note that print mute does not override log level.)

c0d3rman:
(Closed accidentally.)

c0d3rman:
:+1:

PR #74, opened 2016-03-21
Chassis fix

c0d3rman:
Improved Chassis and surrounding classes. Removed obsolete methods, refactored, added and improved comments, and made ChassisMove stop scaling (in favor of doing it in individual Drivers.)


carturn:
:+1: But I still think it should be Cartersian instead of Cartesian

PR #76, opened 2016-03-24
Added motor safety to Motor

carturn:
This fixes a potential issues with the set function in Motors not being called frequently enough.


c0d3rman:
:-1:. In my opinion, forcing motors to be set continuously is unnecessary, restrictive, and doesn't add anything.

arygtm:
This branch uses `MotorSafety` from wpilib to ensure that motors stop if communications are cut.
The robot has proved to be dangerous (especially in the hands of Tobias), but it has never gone crazy because communications were cut. Although this makes the robot slightly safer, I don't think we should merge the branch because constantly setting motors is a bit of a pain.

PR #78, opened 2016-03-25
Changed warnings to debugs where needed.

carturn:


c0d3rman:
:+1:

carturn:
It was decided not to modify chassis as well because there is only one print.

PR #81, opened 2016-03-26
add GuardCommand (useful for AUTO routines)

ajnadel:
This PR adds `GuardCommand`, which allow any indefinite command to be run in parallel with but restricted to the duration of another, set-time command.

### DIFF
--- a/commands/GuardCommand.java
+++ b/commands/GuardCommand.java
@@ -2,6 +2,7 @@ package org.usfirst.frc4904.standard.commands;
 
 
 import edu.wpi.first.wpilibj.command.Command;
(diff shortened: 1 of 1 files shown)


c0d3rman:
`CommandGroup` seems to error when you try to cancel a substituent command. This may not work for `CommandGroup`s (which was its primary purpose).

ajnadel:
@c0d3rman note that `GuardCommand` does not use `CommandGroup`, and a `GuardCommand` won't have to be cancelled -- it will cancel the `parentCommand` which is run separately in `Scheduler` and unaffected by the `GuardCommand`'s caller

c0d3rman:
:+1:

PR #82, opened 2016-03-27
Added SpawnCommand

c0d3rman:
This adds a new command that spawns another command and exits.
This is useful for using commands within commandgroups that usually never end (as this will start them and exit.) This is especially useful for using default commands of subsystems (which usually don't end) inside auton routines.


oshlern:
Could this be replaced with better scheduling (or more options) in CommandGroups?

oshlern:
I now see that while addParallel has the same effect as this within a CommandGroup, the CommandGroup itself when scheduled inside another CommandGroup would never finish.

PR #84, opened 2016-04-04
Autonomous commands

carturn:
Some basic autonomous commands for chassis movement.

### DIFF
--- a/commands/chassis/ChassisMotionProfileDistance.java
+++ b/commands/chassis/ChassisMotionProfileDistance.java
@@ -1,0 +1,69 @@
+package org.usfirst.frc4904.standard.commands.chassis;
+
+
+import org.usfirst.frc4904.standard.LogKitten;
+import org.usfirst.frc4904.standard.custom.ChassisController;
+import org.usfirst.frc4904.standard.custom.motioncontrollers.MotionController;
+import org.usfirst.frc4904.standard.custom.sensors.CustomEncoder;
+import org.usfirst.frc4904.standard.subsystems.chassis.Chassis;
+import edu.wpi.first.wpilibj.command.Command;
+
+public class ChassisMotionProfileDistance extends Command implements ChassisController {
+	private final ChassisMove chassisMove;
+	private final MotionController motionController;
+	private final double distance;
+	private final CustomEncoder[] encoders;
+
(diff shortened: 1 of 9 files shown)


c0d3rman:
This currently drives for a distance and stops, but it is unclear what units the distance is in. It seems not to be inches - it might be pulses, or something else entirely if the conversion was done incorrectly. (Tested using the `encoders` branch of `Offseason2015-A`.)

carturn:
Units are defined by getDistance of the encoder. You change the units with setDistancePerPulse.

c0d3rman:
It seems that the circumference on the Offseason code was incorrect. Also, the initial values were not being subtracted.

c0d3rman:
@carturn should `getDirection` and `getStopped` in `EncoderGroup` reference the underlying encoders? I realize it's hard to synthesize the input, but I think we should just throw an error on disagreement or something, because the current scheme has a couple problems. First, it can report you as stopped if two encoders on the same shaft are inverted or in a similar situation (due to getRate averaging) and also the underlying getStopped doesn't use rate but actually looks at the JNI clock.
Finally, we should have something that warns you whenever the standard deviation (or range) of values reported by encoders in `get`, `getDistance`, and `getRate` gets too big.

c0d3rman:
I am also making `setReverseDirection` invert all encoders instead of just setting them to the same boolean.

carturn:
I agree with changes to setReverseDirection, but I think we should just tell people to use this class for encoders they are confident will display expected behavior for 'get', 'getDistance' and 'getRate'. getStopped and getDirection should reference getRate because getStopped should become true as getRate approaches 0 for good motion control. If your encoders are moving opposite each other, no aspect of this makes sense anyway.

c0d3rman:
But what about the fact that we are now determining the direction of movement via a different method than the underlying controllers?

tylkahn:
This branch deals with fixing the motor controllers so that they are accurate and work properly. This seems like this is something we should keep.

carturn:
This pull request is fully contained within #87. This request is older and is missing a few classes and bug fixes. I think that it should be closed.

carturn:
As this is a duplicate (which I created), it ought to be closed. Please see #87.

PR #87, opened 2016-04-09
Autonomous Commands

carturn:
This is a turn relative to initial robot position.

This includes all autonomous commands, as it uses a couple of them.

Overall: This branch adds the ability to turn absolute and a set distance. It also adds encoderGroups for more convenient use of encoders along with more encoder functions.This is useful for accurate autonomous routines and easier driving.

Adds: ChassisMotionProfileDistance.java, ChassisSetDistance.java, ChassisTurnAbsolute.java, ChassisTurnDegrees.java, EncoderGroup.java. 

ChassisMotionProfileDistance.java: A ChassisController that uses a MotionController to implement motion profiles to drive the robot.

ChassisSetDistance.java: A ChassisConstant that moves the robot a certain distance and speed.

ChassisTurnAbsolute.java: An extension of ChassisTurnDegrees that, instead of turning relative to the starting position, turns relative to the initial position of the robot when turned on.

ChassisTurnDegrees.java: A ChassisController that turns the robot a set amount of degrees relative to the initial angle when starting the command.

EncoderGroup.java: A group of controllers, from which you can get additional things like getAverage().

Also adds a  getDistancePerPulse(), setDistancePerPulse(), getReverseDirection(), and setReverseDirection() for all Encoders.


carturn:
On the previous version of this pull-request, #84, there was a debate regarding what encoder groups should do if the encoders are moving in opposite directions.
We decided that reverse direction should invert each encoder individually, which is a decision I still agree with.
The issue is essentially: if we have two encoders turning in opposite directions, and we try to make an encoder group with them, what should getRate() return? It currently returns the average, but this could be zero if both encoders are moving at the same rate in opposite directions.
I am of the opinion that the average is correct. The context for which it was implemented was to determine how far forwards the chassis has driven. In that case, we want to have 0 if the chassis is only turning in circles (it would be zero since the two sides would be averaged).

oshlern:
It depends on where we use the encoder group, but I don't know if the average is the best measurement for distance traveled. At an instant, this average makes sense for calculating 1d distance traveled (or speed rather), but if we are looking at an extended period of time--as we seem to be in ChassisSetDistance--I don't believe that holds. Summing the instantaneous distances traveled could provide the arc length of the path (not including going back and forth or other similar patterns), but not the displacement. This may be a misunderstanding of how encoders work, and I'm still unsure about the context for using this code. If this isn't gonna be a problem (like when driving straight), then the code seems fine and is labeled as tested, so I would confirm that it is ready for merging.

carturn:
While I agree that the average does not make sense under all circumstances, I am not sure if there is a nicer way to combine the outputs of multiple encoders. There is the case of error detection, but I do not think that there is something we could put into standard to do that easily, e.g. it would depend too much on the configuration of the robot to determine which encoder is most likely to suffer an error. The assumption behind EncoderGroup was that it would only be used when both encoders were expected to be turning in the same direction for the same amount of time.
The intended use case was for dealing with two systems moving in parallel where a small amount of precision could be sacrificed for the sake of ease of use, e.g. driving the chassis in a straight line.
I have added documentation indicating as such.

widdakay:
I think that the encodergroup should average as long as the encoders are reading something that is within some tolerance, and otherwise thrown an error. This way, it is flexible to not be specific to our bot since that is a common way of doing this, and it is easily extensible and has a health check.

oshlern:
I agree. Using simply distance recorded by the encoders (which is nice and simple) there isn't much you can do. When using PID (meaning we expect back and forth motion for correcting movement) then those inaccuracies are mostly undetectable, but we may be able to tell some information from the difference between the distances recorded by encoders if we expect it to be slightly turning at a constant rate. Overall, I'd go with @widdakay 's proposal, which keeps the generalized simplicity while being usable and tolerant in the case of problems.

carturn:
OK. Currently, the throwing of the exception will stop the motion controller or a command based on a sensor.
I have setup the CAN things specifically to throw InvalidSensorExceptions when they do not get packets. I still need to test that the current timeout will generally not drop enough packets to lead to frequent exceptions.

carturn:
Throw for sensor errors revealed that the CAN sensors are dropping really frequently. Polling at 50Hz, I was unable to get data at 1Hz.

carturn:
OK, after a complete redesign of the way we do CAN and the addition of several different types of exception, everything is working. What are everyone's thoughts?

carturn:
I have added places for commands to automatically trigger fallbacks.
At the moment, it is the responsibility of the fallbacks to do cleanup, but I think this ought to be fine.
The idea here is that if the MotorPositionSet fails on a mechanism (like Tim), it will detect this failure, stop PID, then automatically switch to manual control on some joystick axis without requiring a button to be pressed/held.
Note that I am currently not stopping fallbacks in the failing command's end function. I think that in most cases fallbacks will want to stop themselves when appropriate.

c0d3rman:
I meant the other way around - code would be able to continue using `get()` and the like as usual, but implementations that wanted to be safe could call `getSafely()` and catch.

carturn:
I am inclined to think that the default should be to throw exceptions.
I think for almost all cases, excepting resets and maybe excepting certain initializations there should be code in place to handle sensor failure. This would have prevented several issues at Sacramento, SVR, and even Cheezy Champs last year.

c0d3rman:
The issue is that it really overcomplicates simple usage of PID controllers, makes yearly code uglier, and makes it harder to introduce people to PID. If the non-exception-throwing is the default, we are more backwards-compatible, have simpler usage, and still allow for exception catching where it counts (e.g. the standard library). If we made throwing the default, we'd just end up catching and printing the exception everywhere anyway, since for most quick implementations / yearly code there's nothing we can really do with the exception anyway.

carturn:
I do not think there is a case where a motion control system should keep running when a sensor has failed. I think in most cases you should be stopping the motion controller on an exception, then implementing a backup, not just printing. Failure to do so risks incorrect robot behaviour.

c0d3rman:
So we should implement better default behaviour (like returning 0) in the standard library, because if we enforce throwing as default behavior, most people who write PID loops (especially if in a rush) may not even print the exception.

carturn:
That is what it is currently doing within the non-throwing methods, so I agree that return 0 is decent default failure management. However, I am concerned that this will leave us with a lot of commands that are just stuck because they do not know that their associated motion controller has died. The alternative would be having them check if the motion controller has disabled itself, but this would be a less obvious thing to do that catch an exception. I think the no-throw methods should exist in case someone is doing a quick implementation where it does not matter, but I think it should be clear that this may lead to suboptimal behaviour.

c0d3rman:
Agreed, but for the sake of letting the user produce clean, simple code in most cases (where returning 0 is what they would do anyway), I think this should be made clear via the javadoc comment, not via making using the method much more difficult. The point is that if the user hasn't realized the implications of a sensor failing enough to use the safe method and handle it themselves, we'd rather they use our pretty safe default (return 0) than just let eclipse add an empty catch automatically.

carturn:
Code autocomplete is a good point.
We need to make sure that our project settings make Javadoc warnings prominent and start using them more often.
I will change this at some point tonight.

PR #89, opened 2016-06-04
Pause-commands

ave4224:
The purpose of this branch is to allow for commands to be run on a subsystem and resume the command that was running before them upon completion.


commands/InjectedCommand.java
```diff
@@ -0,0 +1,58 @@
...
+	}
+	
+	public InjectedCommand(Command previous, String name, double timeout) {
+		super(name, timeout);
+		this.previous = previous;
+	}
+	
+	@Override
+	final protected void initialize() {
+		if (!previous.isCanceled()) {
```
c0d3rman: This should probably be
```
previous.isRunning() || !previous.isCanceled()
```

commands/InjectedCommandGroup.java
```diff
@@ -0,0 +1,39 @@
...
+	
+	public InjectedCommandGroup(Command previous, String name) {
+		super(name);
+		this.previous = previous;
+	}
+	
+	@Override
+	final protected void interrupted() {
+		onInterrupted();
+		if (!previous.isRunning()) {
```
c0d3rman: Any reason this is not the same check as `InjectedCommand`?

commands/InjectedCommandGroup.java
```diff
@@ -0,0 +1,39 @@
...
+	}
+	
+	@Override
+	final protected void end() {
+		onEnd();
+		if (!previous.isRunning()) {
+			previous.start();
+		}
+	}
+
```
c0d3rman: Any reason you don't override `initialize` like in `InjectedCommand`?

ave4224:
Thanks, All fixed

ave4224:
I added some test cases in both. I don't know how effective they will be, but pretty much what it does is inject a command saying to spin in a circle.
https://github.com/RoboticsTeam4904/2016-Code/tree/power-steering
https://github.com/RoboticsTeam4904/standard/tree/power-steering

c0d3rman:
:+1:

PR #91, opened 2016-09-23
Made CustomPIDController not onTarget right after reset

c0d3rman:
This made GoalAlign immediately stop on the first tick after a reset.


carturn:
:+1:

PR #94, opened 2016-10-15
Identity Modifier should not multiply each value by 1

leijurv:
For one thing, super(1) is parsed as an implicit integer cast to double, which is very slow (at least one cpu instruction).
Also, multiplying the speed by 1.0D each time a motor is set is inefficient (hundreds of extra CPU instructions per second)


c0d3rman:
👍 Looks good to me! Good eye, Leif. Now just needs testing.

c0d3rman:
You can merge this Friday if you test it on the bot, but make sure that a person who didn't write the changes merges it after seeing themselves that it passes the tests. Also change the labels on this PR from "ready for testing" to "tested".

carturn:
:+1:

PR #98, opened 2016-12-07
Threaded Motion Controllers

carturn:
Basically, when we call setPosition on a PostionEncodedMotor, or tell the chassis to rotate to a position, we want it to stay there without a dedicated command calling that function periodically.
This pull request would accomplish this by having motion controllers do their calculation in a separate thread with a constant update frequency (I think 20Hz is optimal, as that is basically the same as the rest of the robot's systems) and push that information to motors.


c0d3rman:
Please drop first cycle to eliminate twitches. (Or constantly calculate D when not enabled)

custom/motioncontrollers/MotionController.java
```diff
@@ -63,14 +89,33 @@ public MotionController(PIDSource source) {
 	 *
 	 * @warning this does not indicate sensor errors
 	 */
-	public abstract void reset();
+	public final void reset() {
+		resetMC();
+		setpoint = sensor.pidGet();
+		synchronized (justReset) {
+			justReset = true;
+		}
```
leijurv: Useless for two reasons: synchronizing on a field that changes doesn't do anything, and setting the value of something doesn't need to be synchronized (what matters for synchronizing is blocks of code that require the value to remain the same)

custom/motioncontrollers/MotionController.java
```diff
@@ -217,4 +282,41 @@ public void disable() {
...
+	 */
+	protected class MotionControllerTask extends TimerTask {
+		@Override
+		public void run() {
+			try {
+				double value = getSafely(); // Always calculate MC output
+				if (justReset) {
+					synchronized (justReset) {
+						justReset = false;
+					}
```
leijurv: Checking the value outside of a synchronized context, then SETTING the value in its own synchronized context????????

custom/motioncontrollers/MotionController.java
```diff
@@ -34,6 +42,9 @@
 	 */
 	public MotionController(PIDSensor sensor) {
 		this.sensor = sensor;
+		output = null;
+		timer = new Timer();
+		task = new MotionControllerTask();
```
ave4224: The reason to use a ScheduledThreadPoolExecutor is to prevent concurrent tasks from delaying each other because Timers use a single thread. However, MotionControllers only have one task each so as long as each has its own instance of a Timer each loop will have its own thread.

Util.java
```diff
@@ -1,5 +1,6 @@
 package org.usfirst.frc4904.standard;
 
+
```
ajnadel: Somebody check this against the formatter.

ave4224:
Ready to Merge. I tested it and it is able to set the wheel to a certain angle. I did not well tune PID but rather did a proof of concept.

Util.java
leijurv: yes that's valid when i had my nested ternary pull request it adde the same newline

commands/motor/MotorPositionSet.java
```diff
@@ -69,6 +67,9 @@ protected void execute() {
 	}
 	
 	@Override
+	protected void execute() {}
```
leijurv: no

Util.java
c0d3rman: Can confirm

c0d3rman:
This is a very large change, and I'd like to get as many people to sign off on it as possible. @leijurv, @ave4224, @ajnadel, @carturn, @widdakay and anyone else, please reread critically - consider edge cases, think about better structure for future maintenance/graceful failure, add as much documentation as possible, etc. Once this has your golden stamp of approval, please approve in a review.

commands/motor/MotorPositionSet.java
carterturn: Yes.
TL;DR: It would be problematic for the motion controller to silently fail; the fallbackCommand should be started on sensor failure, regardless of whether it was during initialization or while running.
Given that the way some errors are detected is through timeouts, it is actually more likely that an error will become apparent while running than during initialization. While the motion controller will disable itself either way, it does not have any ability to trigger the fallback without the command actively checking for errors. It might be possible to implement that, but I am inclined to prefer to separate the fallback command from the controller.

carterturn:
I think it would be nice to make the ChassisControllers be able to use the setOutput of the motion controller, as this would allow the movement to be slightly more precise.
Also, I think we need to either ensure there is a standard for motion controllers being enabled and disabled or add the relevant enable and disable calls to all commands. At the moment, there are a few commands that either enable or disable a motion controller. They do not appear to restore to the original state after running, which is suboptimal. I think it would be safer to have the initialize in all core motor and chassis commands set the motion controller to enabled or disabled, especially as with setOutput the ending of a command need not necessarily trigger the ending of a motion control.

ave4224:
It might be useful to include [this](http://pastebin.com/DbHFGmkL), but I don't know how I would make a test case.

carterturn:
@ave4224 I think it would be better to just call getSafely on the motion controller. There is not really an advantage to having the motion controller write to a class, then reading a variable from that class.

ave4224:
That makes sense.
On Jan 23, 2017, at 6:42 PM, Carter Turnbaugh <notifications@github.com<mailto:notifications@github.com>> wrote:


@ave4224<https://github.com/ave4224> I think it would be better to just call getSafely on the motion controller. There is not really an advantage to having the motion controller write to a class, then reading a variable from that class.

custom/MCChassisController.java
```diff
@@ -8,15 +8,15 @@
...
+public class MCChassisController implements ChassisController {
 	protected ChassisController controller;
 	protected double maxDegreesPerSecond;
 	protected double targetYaw;
 	protected double lastUpdate;
 	protected IMU imu;
 	protected MotionController motionController;
 	
-	public PIDChassisController(ChassisController controller, IMU imu, MotionController motionController, double maxDegreesPerSecond) {
+	public MCChassisController(ChassisController controller, IMU imu, MotionController motionController, double maxDegreesPerSecond) {
```
ajnadel: The official pronunciation of this class is "mick chassis controller" as if it were a McDonalds product.

ajnadel:
Tested!

PR #99, opened 2016-12-09
CommandRobotBase cleanup

carturn:
Adds a function that cleans up all teleop, autonomous, and healthcheck commands when any new state is initialized.


CommandRobotBase.java
```diff
@@ -36,7 +36,28 @@ protected final void displayChoosers() {
...
+	 * with the current robot mode.
+	 */
+	private void cleanup() {
+		if (autonomousCommand != null) {
+			autonomousCommand.cancel();
+		}
+		if (teleopCommand != null) {
+			teleopCommand.cancel();
+		}
+		if (healthcheckCommand != null) {
```
HowardStark: Are we going to be implementing health checks this year?

PR #101, opened 2017-01-10
Twitchtimv

leijurv:


leijurv:
@c0d3rman

leijurv:
the reason why it's in sensor motor is that PIDController.get has to return *some* number, not like it can return null (hmm). The issue is that it doesn't know what to return at all on the first tick, so it's returning 0 or similar, *that's what causing the twitch*. Whatever's calling PIDController I suppose would need to detect the first iteration and ignore the result, that's why I made justReset public and implemented it in sensor motor

leijurv:
Maybe PIDController should return an Optional<Double> or a @Nullable Double

c0d3rman:
That sounds like a good idea - perhaps add a second `getNullable` method (or a better name) so if you're just making a quick implementation you can use `get` but implementations in the standard library (like in SensorMotor) can use `getNullable`? That sounds great – if you think it's worth adding while we get the thread implementation working go for it. (If not it may get incorporated into that implementation if needed)

PR #103, opened 2017-01-19
added minimum nominal output functionality to custom pid controller

blobfish2000:
Added minNominalOutput to CustomPIDController.
added setMinNominalOutput() to CustomPIDController, sets the minNominalOutput.
added getMinNominalOutput() to CustomPIDController, gets the minNominalOutput.
Edited get() so that if |result| is less than the minNominalOutput than it puts it to minNominalOutput or -minNominalOutput depending on if it's above or below 0.


c0d3rman:
I think that it's fine to add this to the base CustomPIDController - we can make the class feature-rich with the understanding some features won't always be used.

c0d3rman:
We should highlight that this feature only makes sense if the default output for the motor controller is 0, such as when controlling a motor or a chassis' turn speed.
Also, what do we want to do if a minimum nominal output is set without setting a tolerance? (In most cases this would cause the PID loop to never complete.) I think we shouldn't go so far as to throw an exception, as the code may set the minimum nominal output first and set the tolerance immediately after. Maybe a `LogKitten.w()`?

PR #104, opened 2017-01-19
better date format string

leijurv:


leijurv:
Note: old code would generate `2017-1-19_15:54:42`, new code generates `2017-01-19_15:54:42`. Old code was fine and generally generated constant length dates, because build season is in the earlier part of the year (not months 10 through 12)

leijurv:
here's the test case I used to ensure behavior was the same:
```
package wewlad;
import java.text.SimpleDateFormat;
import java.util.Calendar;
import java.util.Date;
public class Wewlad {
public static void main(String[] args){
	Calendar now = Calendar.getInstance();
	String timestamp = Integer.toString(now.get(Calendar.YEAR));
	timestamp += "-" + Integer.toString(now.get(Calendar.MONTH) + 1);
	timestamp += "-" + Integer.toString(now.get(Calendar.DATE));
	timestamp += "_" + Integer.toString(now.get(Calendar.HOUR_OF_DAY));
	timestamp += ":" + Integer.toString(now.get(Calendar.MINUTE));
	timestamp += ":" + Integer.toString(now.get(Calendar.SECOND));
	System.out.println(timestamp);
	String dateString = null;
	   SimpleDateFormat sdfr = new SimpleDateFormat("yyyy-M-dd_HH:mm:ss");
	   try{
		dateString = sdfr.format( new Date() );
	   }catch (Exception ex ){
		System.out.println(ex);
	   }
	   System.out.println(dateString);
}
}
```

LogKitten.java
```diff
@@ -359,14 +360,8 @@ public static synchronized void clean() {
...
 	private static String timestamp() {
-		Calendar now = Calendar.getInstance();
-		String timestamp = Integer.toString(now.get(Calendar.YEAR));
-		timestamp += "-" + Integer.toString(now.get(Calendar.MONTH) + 1);
-		timestamp += "-" + Integer.toString(now.get(Calendar.DATE));
-		timestamp += "_" + Integer.toString(now.get(Calendar.HOUR_OF_DAY));
-		timestamp += ":" + Integer.toString(now.get(Calendar.MINUTE));
-		timestamp += ":" + Integer.toString(now.get(Calendar.SECOND));
-		return timestamp;
+		SimpleDateFormat sdfr = new SimpleDateFormat("yyyy-MM-dd_HH:mm:ss");
```
c0d3rman: The date string format should probably be a protected static constant. Also do we need the temporary variable? (This method is called many times a tick.)

leijurv:
@c0d3rman fixt

LogKitten.java
```diff
@@ -360,8 +361,7 @@ public static synchronized void clean() {
 	 * @return timestamp as string in the format "YEAR-MONTH-DAY_HOUR:MIN:SEC"
 	 */
 	private static String timestamp() {
-		SimpleDateFormat sdfr = new SimpleDateFormat("yyyy-MM-dd_HH:mm:ss");
-		return sdfr.format(new Date());
+		return TIMESTAMP_FORMAT.format(new Date());
```
c0d3rman: By convention, this should be accessed statically. (Eclipse auto-formats this - make sure to use eclipse or run your code by the formatter.)

leijurv:
@c0d3rman fixt

leijurv:
Piggybacker

PR #105, opened 2017-01-20
rethrow exceptions instead of just printing stack trace and continuing

leijurv:


leijurv:
Why wouldn't this be correctable?

leijurv:
Throwing exceptions when an error occurs is a good idea even when the code cannot recover from said error (such as the "this should never happen" error)

carturn:
Is there a way to catch an error from a static initializer? If so, then this would be fine supposing that we did so.

carturn:
In the case of logging, it is more important that the rest of the robot work than logging work. The worst-case scenario would be code failure for the duration of a match, and while this is unlikely, it should be avoided. I think in the case of the robot, if an error can not be recovered from, the subsystem the error occurred within should be disabled, but the rest of the code should simply ignore the error.

leijurv:
Ah, I see.

PR #107, opened 2017-01-20
kitten level as enum instead of class with public constructor

leijurv:


leijurv:
The backwards compatibility (in which referencing the field LogKitten.LEVEL_whatever still works) should help

leijurv:
One thing to be careful of: I had to remove compareTo because Enum.compareTo is final and can't be overridden. Comparisons should be the same, as the enum values are listed in the correct order, but this is a coincidence at the moment.

c0d3rman:
This looks like an excellent addition. @leijurv, perhaps add a comment indicating the importance of declaring KittenLevels in order? Also, I would appreciate if you could update the title/description of this PR to better reflect what the changes are and why you made them (for posterity).

leijurv:
@c0d3rman thoughts on severity as the ordinal of the enum?

c0d3rman:
@leijurv I think it's very clever! I would add a clarifying comment, though.

leijurv:
It does have `// Defined in decreasing order of severity. Enum.compareTo uses the definition order to compare enum values.`

c0d3rman:
Yes, which is why I approved the changes as is - I just thought it'd be a nice addition. Thanks for adding it! I think this is 100% ready for testing.

HowardStark:
Tested. Looks good.

PR #108, opened 2017-01-20
Flipped getDistance and getDistanceSafely

c0d3rman:
Corrected a bug in standard where DistanceSensor's `getDistance` and `getDistanceSafely` have the opposite meanings to the corresponding methods in PIDSensor


ajnadel:
+1

PR #109, opened 2017-01-22
replaced every instanceof string with object in logkitten

leijurv:
Yes that would require a new function signature for every primitive. However, thanks to Java autoboxing int -> Integer and such, you can pass an int literal to a function that expects an Object. So, LogKitten.wtf(5) will work already for example. @c0d3rman

### DIFF
--- a/LogKitten.java
+++ b/LogKitten.java
@@ -147,7 +147,8 @@ public class LogKitten {
 		HAL.sendError(true, logLevel.getSeverity(), false, errorMessage, details, "", false);
 	}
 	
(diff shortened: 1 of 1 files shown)


c0d3rman:
Looks good! Is there any way to integrate accepting primitives (e.g. `double`) into this, or does that require a new function signature for every primitive? @leijurv

leijurv:
Yes that would require a new function signature for every primitive. However, thanks to Java autoboxing int -> Integer and such, you can pass an int literal to a function that expects an Object. So, `LogKitten.wtf(5)` will work already for example.

PR #111, opened 2017-01-22
kittenlevel doesn't need to implement comparator, only comparable

leijurv:
> No one needs to call LEVEL_DEBUG.compare(LEVEL_LOG,LEBEL_VERBOSE); all that's needed is LEVEL_DEBUG.compareTo(LEVEL_LOG), which is comparable.

### DIFF
--- a/LogKitten.java
+++ b/LogKitten.java
@@ -6,7 +6,6 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.text.SimpleDateFormat;
-import java.util.Comparator;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


c0d3rman:
Can you explain why "kittenlevel doesn't need to implement comparator, only comparable" in the description so I can approve? @leijurv

leijurv:
No one needs to call LEVEL_DEBUG.compare(LEVEL_LOG,LEBEL_VERBOSE); all that's needed is LEVEL_DEBUG.compareTo(LEVEL_LOG), which is comparable.

c0d3rman:
Ah, that makes sense. Approved!

ajnadel:
Tested +1

PR #114, opened 2017-01-22
crucial performance optimization

leijurv:
crucial performance optimization


ajnadel:
This reduces readability for a trivial performance boost.

PR #115, opened 2017-01-22
streams

leijurv:


c0d3rman:
This was closed because streams provide no real benefits in this scenario but are unintuitive to beginners.

PR #117, opened 2017-01-22
remove epsilon

leijurv:
Replace epsilon with double min value

### DIFF
--- a/Util.java
+++ b/Util.java
@@ -15,7 +15,7 @@ public class Util {
 	 * @param var
 	 *        The floating point number you want to compare
 	 * @return
-	 * 		Whether or not it is within Constants.EPSILON of zero
+	 * 		Whether or not it is within Double.MIN_VALUE of zero
 	 */
 	public static boolean isZero(double var) {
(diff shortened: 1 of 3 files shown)


Util.java
```diff
@@ -1,16 +1,12 @@
...
 
 /**
  * Common utilities
  */
 public class Util {
-	/**
-	 * A constant for dealing with floating point errors.
-	 * Add/subtract this from floats when doing equality comparisons.
-	 */
-	public static final double EPSILON = 0.0000001;
```
c0d3rman: Please remove this line

ajnadel: @c0d3rman Isn't that line removed?

Util.java
```diff
@@ -22,7 +18,7 @@
 	 * 		Whether or not it is within Constants.EPSILON of zero
```
ajnadel: @leijurv Please adjust this line as well.

leijurv: k

c0d3rman:
@ajnadel, your requested change has been made

PR #121, opened 2017-01-24
DateFormats aren't thread safe in java

leijurv:
http://stackoverflow.com/questions/4021151/java-dateformat-is-not-threadsafe-what-does-this-leads-to


c0d3rman:
Good catch. Since many (all) threads will be using `DateFormat` many times a second, perhaps we want to go with the `ThreadLocal` approach from that link? @leijurv

carturn:
I am inclined to agree with the ThreadLocal approach, it would not be good for motion controller threads to block due to logging.
By the way, I think it is worth mentioning that prior to motion controller threads, we did not actually have any other threads (I think), so that is probably why we did not observe this error.

leijurv:
Not sure what you're talking about: LogKitten.logMessage is *already* synchronized...

leijurv:
This won't add anything new, because when you call a synchronized method from another synchronized method, it doesn't need to aquire the lock (the thread already has it)

carturn:
That is true. We should work out a way to get around the synchronization (maybe a queue, but that seems bad), or ensure that we remove logging from motion controller threads, as the timing for motion controller threads is supposed to be consistent.

leijurv:
```
int possibles=0;
	SimpleDateFormat sdfr = new SimpleDateFormat("yyyy-M-dd_HH:mm:ss");
	String dateString = null;
	   
	while (System.currentTimeMillis() - start < 1000) {
		possibles++;
		dateString = sdfr.format( new Date() );
	}
	System.out.println("in 1000ms able to increment an int "+possibles+" times");
	Calendar now = Calendar.getInstance();
```

Simple benchmarks indicate that `dateFormat.format(new Date());` is fast enough to run 912,716 per second. Synchronizing on it will not be an issue (especially since logMessage is already synchronized, this won't have any affect on performance as of the current state of LogKitten)

ajnadel:
Tested on Pinky.

PR #123, opened 2017-01-24
These are never modified therefore can be final

leijurv:
While `logLevel`, `dsLevel`, and `printLevel` are modified, `DEFAULT_LOG_LEVEL` and co aren't


c0d3rman:
Huh. I could have sworn those were final.

PR #125, opened 2017-01-24
Use stringbuilder instead of appending to string in loop

leijurv:


commands/chassis/ChassisMove.java
```diff
@@ -76,14 +76,16 @@ protected void initialize() {
...
 		motorSpeeds = chassis.getMotorSpeeds();
-		String motorSpeedsString = "";
+		StringBuilder motorSpeedsString = new StringBuilder();
+		motorSpeedsString.append("Motor speeds: ");
 		for (int i = 0; i < motorSpins.length; i++) {
 			LogKitten.d(Double.toString(motorSpeeds[i]));
 			motorSpins[i].set(motorSpeeds[i]);
-			motorSpeedsString += Double.toString(motorSpeeds[i]) + " ";
+			motorSpeedsString.append(motorSpeeds[i]);
+			motorSpeedsString.append(' ');
```
c0d3rman: To prevent a trailing space on the string, I would move this before the previous line and remove the trailing space  on line 80.

leijurv: Okay. The previous behavior was a trailing space though

PR #126, opened 2017-01-24
Inner classes should be static wherever possible

leijurv:


leijurv:
I'm not sure if there's a precedent for accessing static inner classes, but I would lean towards accessing it as it is now. Non-static inner classes are bad practice (e.g. because they maintain a reference to their enclosing instance which hurts gc)

c0d3rman:
Does it pass the formatter?
Also, I agree with the idea of static inner classes, I'm just trying to decide on a convention on how to access them.

leijurv:
closing in favor of #129

c0d3rman:
Please make sure to delete branches on closed PRs @leijurv

PR #127, opened 2017-01-24
use entry set iterators wherever possible

leijurv:


c0d3rman:
@leijurv, could you update the PR with an explanation of this change? It presents a significant reduction to readability, so unless it has some benefit I don't think it's necessary.

leijurv:
Does it? Instead of iterating over the keys and calling map.get(key) repeatedly, it iterates over the entries in the map (which is a key value pair)

PR #128, opened 2017-01-24
Fix CAN Read Issue

carturn:
This is to fix an ongoing issue with CAN reading and implement a better sensor management system upon that being fixed.
According to typical CAN system designs, sensors should be pushing data constantly. Previously, we were using a call and response system.
There was previously an error where, due to incorrectly initialized CAN ids on the Teensy side, most messages were being lost. In addition, the mask on the RoboRIO side had an error that was leading to further message loss, although this was introduced more recently.

The RoboRIO masking issues are fixed here, and the previous measures taken to allow some communication were eliminated. Sensors no longer attempt to write before reading either.

### DIFF
--- a/custom/CustomCAN.java
+++ b/custom/CustomCAN.java
@@ -15,7 +15,7 @@ public class CustomCAN {
 	// Because CANJNI is basically static, we do not extend it.
 	protected final int messageID;
 	protected final String name;
-
+	
 	/**
 	 * Constructor for a CustomCAN device.
 	 * The name is local and for your convenience only.
@@ -29,11 +29,11 @@ public class CustomCAN {
 		this.name = name;
 		messageID = 0x00000FFF & id; // Ensure that the messageID is zeroed (32 bit int should be default, but better to be careful)
 	}
-
+	
 	public String getName() {
 		return name;
 	}
-	
+
 	/**
 	 * Used to write data to the device.
 	 *
@@ -43,12 +43,10 @@ public class CustomCAN {
 	 */
 	public void write(byte[] data) {
 		ByteBuffer canData = ByteBuffer.allocateDirect(8);
-		for (int i = 0; i < 8; i++) {
-			canData.put(i, data[i]);
-		}
+		canData.put(data);
 		CANJNI.FRCNetCommCANSessionMuxSendMessage(messageID, canData, CANJNI.CAN_SEND_PERIOD_NO_REPEAT);
 	}
-	
+
 	/**
 	 * Read data as bytebuffer
 	 *
@@ -71,7 +69,7 @@ public class CustomCAN {
 		}
 		return response;
 	}
-	
+
 	/**
 	 * Reads data
 	 * Also stops repeating the last message.
--- a/custom/CustomCAN.java
+++ b/custom/CustomCAN.java
@@ -15,7 +15,7 @@ public class CustomCAN {
 	// Because CANJNI is basically static, we do not extend it.
 	protected final int messageID;
 	protected final String name;
-	
+
 	/**
 	 * Constructor for a CustomCAN device.
 	 * The name is local and for your convenience only.
@@ -27,16 +27,15 @@ public class CustomCAN {
 	 */
 	public CustomCAN(String name, int id) {
 		this.name = name;
-		messageID = id;
+		messageID = 0x00000FFF & id; // Ensure that the messageID is zeroed (32 bit int should be default, but better to be careful)
 	}
-	
+
 	public String getName() {
 		return name;
 	}
-
+	
 	/**
 	 * Used to write data to the device.
-	 * Will continue writing until read is called.
 	 *
 	 * @param data
 	 *        Data to be written. Should be EXACTLY 8 bytes long ONLY.
@@ -49,7 +48,13 @@ public class CustomCAN {
 		}
 		CANJNI.FRCNetCommCANSessionMuxSendMessage(messageID, canData, CANJNI.CAN_SEND_PERIOD_NO_REPEAT);
 	}
-
+	
+	/**
+	 * Read data as bytebuffer
+	 *
+	 * @return
+	 * 		ByteBuffer containing CAN message, or null
+	 */
 	protected ByteBuffer readBuffer() {
 		IntBuffer idBuffer = ByteBuffer.allocateDirect(4).asIntBuffer();
 		idBuffer.clear();
@@ -66,7 +71,7 @@ public class CustomCAN {
 		}
 		return response;
 	}
-
+	
 	/**
 	 * Reads data
 	 * Also stops repeating the last message.
--- a/custom/CustomCAN.java
+++ b/custom/CustomCAN.java
@@ -28,7 +28,7 @@ public class CustomCAN {
 	 */
 	public CustomCAN(String name, int id) {
 		this.name = name;
-		messageID = 0x00000FFF & id; // Ensure that the messageID is zeroed (32 bit int should be default, but better to be careful)
+		messageID = id; // Ensure that the messageID is zeroed (32 bit int should be default, but better to be careful)
 	}
 
 	public String getName() {
@@ -63,7 +63,7 @@ public class CustomCAN {
 		long start = System.currentTimeMillis();
 		while (System.currentTimeMillis() - start < CustomCAN.CAN_MAX_READ_WAIT) {
 			try {
-				response = CANJNI.FRCNetCommCANSessionMuxReceiveMessage(idBuffer, 0x0fffffff, timestamp);
+				response = CANJNI.FRCNetCommCANSessionMuxReceiveMessage(idBuffer, 0x1fffffff, timestamp);
 				break;
 			}
 			catch (CANMessageNotFoundException e) {}
--- a/custom/CustomCAN.java
+++ b/custom/CustomCAN.java
@@ -15,6 +15,7 @@ public class CustomCAN {
 	// Because CANJNI is basically static, we do not extend it.
 	protected final int messageID;
 	protected final String name;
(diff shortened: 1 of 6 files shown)


carturn:
This code has been tested outside of Standard, but I would like to test this on Pinky before merging.

carturn:
OK, there is now a change to how we handle CAN sensors in general that is worth noting.
With this pull request, sensors will return two integers, rather than one integer and a meta-integer.
Based on reading Talon related code, different message IDs should be used for metadata. I think this should be the plan in the future, since there is little to no chance of us needing more than 32 separate sensors (which would allow us to allocate each sensor 8 message IDs, more if most of our sensors are just encoders and distance sensors).

custom/CustomCAN.java
```diff
@@ -36,36 +36,37 @@ public String getName() {
...
+	 */
 	protected ByteBuffer readBuffer() {
 		IntBuffer idBuffer = ByteBuffer.allocateDirect(4).asIntBuffer();
 		idBuffer.clear();
 		idBuffer.put(0, Integer.reverseBytes(messageID));
 		ByteBuffer timestamp = ByteBuffer.allocate(4);
 		ByteBuffer response = null;
 		long start = System.currentTimeMillis();
-		while (System.currentTimeMillis() - start < 10) {
+		while (System.currentTimeMillis() - start < 5) {
```
c0d3rman: Perhaps this should be a constant field instead of a magic number?

custom/sensors/CANEncoder.java
```diff
@@ -144,7 +144,6 @@ public boolean getStoppedSafely() throws InvalidSensorException {
 	@Override
 	public void reset() {
 		super.write(new byte[] {0x72, 0x65, 0x73, 0x65, 0x74, 0x65, 0x6e, 0x63}); // resetenc
```
c0d3rman: Again, perhaps this byte sequence should be a constant somewhere? (I don't know enough about what it actually is to say one way or the other)

custom/sensors/CANSensor.java
```diff
@@ -11,89 +11,49 @@
  */
 public class CANSensor extends CustomCAN {
 	private final int[] values;
-	private final long[] ages;
-	private final long MAX_AGE = 1000;
+	private long age; // data age
+	private final long MAX_AGE = 100;
```
c0d3rman: Should this be static?

custom/sensors/CANSensor.java
```diff
@@ -11,89 +11,49 @@
...
-				return values[mode];
-			}
+			values[0] = (int) data & 0xFFFFFFFF;
+			values[1] = (int) (data >> 32) & 0xFFFFFFFF;
+			age = System.currentTimeMillis();
+			return values;
 		}
-		if (System.currentTimeMillis() - ages[mode] > MAX_AGE) {
-			throw new InvalidSensorException("CAN data oudated For CAN sensor " + getName() + " with ID " + messageID);
+		if (System.currentTimeMillis() - age > MAX_AGE) {
```
c0d3rman: The name `age` is misleading, especially since `MAX_AGE` is not the maximum value of `age`.

custom/sensors/CANSensor.java
```diff
@@ -11,89 +11,49 @@
...
+	 *         If the available data is more than one tenth of a second old,
 	 *         this function will throw an InvalidSensorException
 	 *         to indicate that.
 	 */
-	public int read(int mode) throws InvalidSensorException {
-		write(new byte[] {(byte) ((byte) mode & 0xFF), 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01}); // Write to trigger read
-		ByteBuffer rawData = readBuffer();
+	public int[] readSensor() throws InvalidSensorException {
+		ByteBuffer rawData = super.readBuffer();
 		if (rawData != null && rawData.remaining() > 7) {
```
c0d3rman: Where did this 7 come from?

custom/sensors/CANSensor.java
carturn: Yes, will change and test tomorrow morning.

custom/sensors/CANSensor.java
carturn: A message is 8 bytes, so a real message is greater than 7 bytes.
This is part of the CAN standard.

custom/sensors/CANSensor.java
carturn: True, I will change and test tomorrow morning.

custom/CustomCAN.java
carturn: Probably. I will change and test tomorrow morning.

custom/sensors/CANSensor.java
c0d3rman: OK

carturn:
OK, requested changes have been made and it still works (not too surprising).

custom/CustomCAN.java
```diff
@@ -15,6 +15,7 @@
 	// Because CANJNI is basically static, we do not extend it.
 	protected final int messageID;
 	protected final String name;
+	private static final int CAN_MAX_READ_WAIT = 5; // How long to wait for a CAN message before returning null
```
c0d3rman: Mention that this is milliseconds

custom/sensors/CANSensor.java
```diff
@@ -42,15 +42,15 @@ public CANSensor(String name, int id) {
 	 */
 	public int[] readSensor() throws InvalidSensorException {
 		ByteBuffer rawData = super.readBuffer();
-		if (rawData != null && rawData.remaining() > 7) {
+		if (rawData != null && rawData.remaining() > 7) { // 8 is minimum CAN message length
```
c0d3rman: Perhaps this should be >= 8 to better reflect what's happening?

custom/CustomCAN.java
carturn: Done. Also done where there was another problem in a similar file.

custom/sensors/CANEncoder.java
```diff
@@ -14,9 +14,14 @@
 	private PIDSourceType pidSource;
 	private double distancePerPulse;
 	private boolean reverseDirection;
+	/**
+	 * Sequence of bytes used to reset an encoder
+	 * Spells out "resetenc" in ASCII
+	 */
+	private static final byte[] RESET_ENCODER_BYTE_SEQUENCE = new byte[] {0x72, 0x65, 0x73, 0x65, 0x74, 0x65, 0x6e, 0x63};
```
leijurv: Perhaps `private static final byte[] RESET_ENCODER_BYTE_SEQUENCE = "resetenc".getBytes();`

custom/sensors/CANEncoder.java
```diff
@@ -143,8 +148,7 @@ public boolean getStoppedSafely() throws InvalidSensorException {
 	 */
 	@Override
 	public void reset() {
-		super.write(new byte[] {0x72, 0x65, 0x73, 0x65, 0x74, 0x65, 0x6e, 0x63}); // resetenc
-		super.read(); // and stop transmitting resets once the teensy replies
```
leijurv: why the removal of this line

custom/sensors/CANUltrasonicDistanceSensor.java
```diff
@@ -17,20 +15,20 @@
...
 			LogKitten.ex(e);
 			return 0;
 		}
 	}
-	
+
 	@Override
 	public double getDistanceSafely() throws InvalidSensorException {
-		return super.read(CANUltrasonicDistanceSensor.CAN_SENSOR_MODE);
+		return super.readSensor()[0];
```
leijurv: why is CAN_SENSOR_MODE no longer a constant? what previously was `super.read(CANUltrasonicDistanceSensor.CAN_SENSOR_MODE);` could become `super.readSensor()[CANUltrasonicDistanceSensor.CAN_SENSOR_MODE];`, no?

custom/sensors/CANSensor.java
```diff
@@ -11,89 +11,49 @@
...
-				return values[mode];
-			}
+			values[0] = (int) data & 0xFFFFFFFF;
+			values[1] = (int) (data >> 32) & 0xFFFFFFFF;
+			lastRead = System.currentTimeMillis();
+			return values;
 		}
-		if (System.currentTimeMillis() - ages[mode] > MAX_AGE) {
-			throw new InvalidSensorException("CAN data oudated For CAN sensor " + getName() + " with ID " + messageID);
+		if (System.currentTimeMillis() - lastRead > CANSensor.MAX_AGE) {
```
leijurv: Why removing the per-mode age?

carturn: There are no more modes; one packet contains all of the data, so all values have the same age.

custom/sensors/CANEncoder.java
carturn: That seems good, I will add it.

custom/sensors/CANEncoder.java
carturn: Because we no longer need to read after writing.

carturn:
OK, Leif's comments have been addressed.

custom/sensors/CANUltrasonicDistanceSensor.java
c0d3rman: I think this is also answered by @carturn's comment:

> There are no more modes; one packet contains all of the data, so all values have the same age.

c0d3rman:
IDK if the distance sensor array constants are really needed, but whatever. Is this tested? If so I'll merge.

carturn:
This is tested, but should wait for merging until [FRCTeensyCAN #12](https://github.com/RoboticsTeam4904/FRCTeensyCAN/pull/12), as message formats have changed.

PR #131, opened 2017-01-27
fixed generics in typednamedsendablechooser

leijurv:
like the other generic pull, this one fixes all the eclipse warnings for the file, etc


c0d3rman:
@leijurv Could you document how to enable these warnings for future people reviewing the standard library?

leijurv:
I mean... the warnings are no longer being shown because they are fixed... To replace the errors, undo this change

c0d3rman:
No, I mean when they are investigating other files, how to see all Eclipse warnings (the same ones you used to find this issue). Or are all warnings shown by default in the Problems pane?

leijurv:
Oh, I have FindBugs and PMD installed. But the generic type-warnings I was talking about are from javac / eclipse warnings.

PR #132, opened 2017-01-27
better string joining

leijurv:
i am amazed by all the inventive ways to concatenate strings in the standard library. in this case, we shouldn't create a new string array for no reason. (String.join uses the same StringJoiner internally, so this can only help not hurt)

### DIFF
--- a/commands/RunAll.java
+++ b/commands/RunAll.java
@@ -1,6 +1,7 @@
 package org.usfirst.frc4904.standard.commands;
 
 
+import java.util.StringJoiner;
 import edu.wpi.first.wpilibj.command.Command;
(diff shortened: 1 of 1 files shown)


commands/RunAll.java
```diff
@@ -10,10 +11,10 @@ public RunAll(String name) {
 	}
 	
 	protected static String joinNames(Command... commands) {
-		String[] names = new String[commands.length];
-		for (int i = 0; i < commands.length; i++) {
-			names[i] = commands[i].getName();
+		StringJoiner joiner = new StringJoiner(", ");
+		for (Command cs : commands) {
```
c0d3rman: Why `cs`?

PR #134, opened 2017-01-27
rearranged constructors and finalized field

leijurv:
also removed the constructor that results in `state` being null, that can only end in a nullpointerexception somewhere in a switch


commands/chassis/ChassisShift.java
```diff
@@ -10,44 +10,25 @@
  */
 public class ChassisShift extends Command {
 	protected final SolenoidShifters solenoids;
-	protected SolenoidShifters.ShiftState state;
+	protected final SolenoidShifters.ShiftState state;
 	
 	/**
-	 * Shifts the solenoids to the opposite state
+	 * Shifts the solenoids to the state state
```
c0d3rman: This isn't exactly a helpful comment.

commands/chassis/ChassisShift.java
```diff
@@ -10,44 +10,25 @@
...
 	 *
 	 * @param solenoids
+	 * @param state
 	 */
-	public ChassisShift(SolenoidShifters solenoids) {
+	public ChassisShift(SolenoidShifters solenoids, SolenoidShifters.ShiftState state) {
```
c0d3rman: Why did we change this command from a toggle?

c0d3rman: Also, this is a breaking change for anything that uses a shifting chassis.

leijurv: Actually, anything that called the constructor `public ChassisShift(SolenoidShifters solenoids) {` would set shift to null, and cause a nullpointerexception on the first switch for it

c0d3rman: I see. Regardless, why did we remove the toggle behavior rather than fix it?

commands/chassis/ChassisShift.java
leijurv: that was unintentional

commands/chassis/ChassisShift.java
leijurv: idk

commands/chassis/ChassisShift.java
```diff
@@ -13,7 +13,7 @@
 	protected final SolenoidShifters.ShiftState state;
 	
 	/**
-	 * Shifts the solenoids to the state state
+	 * Shifts the solenoids to the opposite state
```
c0d3rman: I would recommend that this comment say something like `Shifts the solenoids to the given state` and that the one on the toggle constructor says something along the lines of `Toggles the solenoids`

leijurv:
@c0d3rman

PR #141, opened 2017-01-27
thread sleep in loop

leijurv:
@carterturn @carturn 
what are your thoughts on this? 
a simple test on my laptop

```
	int possibles=0;
	while (System.currentTimeMillis() - start < 10) {
		possibles++;
	}
	System.out.println("in 10ms able to increment an int "+possibles+" times");
```

indicates that a 10ms waiting while loop can run up to 100,000 times. This is because a loop like this uses 100% of a CPU core just running over and over. This pull reduces the polling rate for CAN from ~10,000,000 times a second to 1,000 times a second, by waiting 1ms between each loop iteration.

### DIFF
--- a/custom/CustomCAN.java
+++ b/custom/CustomCAN.java
@@ -33,7 +33,7 @@ public class CustomCAN {
 	public String getName() {
 		return name;
 	}
-
+	
 	/**
 	 * Used to write data to the device.
 	 * Will continue writing until read is called.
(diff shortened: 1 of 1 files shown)


carturn:
Based on recent tests, there is almost no queue for CAN messages, so I am concerned that 1ms might be too long (especially as it could delay for longer than that).
That being said, some smaller delay would probably be nice.
It is probably not as bad as the test implies though, as CANJNI.FRCNetCommCANSessionMuxReceiveMessage does involve waiting for a hardware response.

carturn:
So, I read through the PDP code, available [here](https://github.com/wpilibsuite/allwpilib/blob/master/hal/lib/athena/ctre/PDP.cpp). Based on this, it should generally be possible to receive messages only calling the function once.
I think it would be worthwhile to see if we can avoid having a loop around CANJNI.FRCNetCommCANSessionMuxReceiveMessage by increasing Teensy update rate.

carturn:
Quick tests indicate that both the 1ms sleep and the non-loop have no apparent impact on performance with normal robot code.
I would like to test the limits of all approaches before declaring this entirely tested though.

carturn:
OK, better test conducted. We do not actually seem to need the loop anymore at all.

carturn:
While this works, I think #153 is cleaner.

PR #145, opened 2017-01-27
Replace Encoder Group with Encoder Pair

carturn:
Encoder Group seems to have limited use cases and limitations due to it using N-encoders.
Therefore, I am proposing replacing it with a pair of encoders, which I think covers most of our uses for multiple encoders together. I think this will simplify implementation of encoder based chassis driving systems.


carturn:
@ave4224 would this help with the encoder based chassis move you were working on?

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+		}
+		catch (Exception e) {
+			LogKitten.ex(e);
+			return 0;
+		}
+	}
+	
+	@Override
+	public boolean getDirection() {
+		return (getRate() > 0);
```
leijurv: Parens.

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+	}
+	
+	@Override
+	public boolean getDirectionSafely() throws InvalidSensorException {
```
leijurv: Parens

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+	public double getRateSafely() throws InvalidSensorException {
+		return (encoders[0].getRateSafely() + encoders[1].getRateSafely()) / 2.0;
+	}
+	
+	@Override
+	public double getRate() {
+		try {
+			return getRateSafely();
+		}
+		catch (Exception e) {
```
leijurv: Catch the specific checked exception not a general exception. Or use | syntax

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+	public int getSafely() throws InvalidSensorException {
+		return (int) Math.round(((double) (encoders[0].getSafely() - offset[0]) + (double) (encoders[1].getSafely() - offset[1])) / 2.0);
+	}
+	
+	@Override
+	public int get() {
+		try {
+			return getSafely();
+		}
+		catch (Exception e) {
```
leijurv: Same here

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+public class EncoderPair implements CustomEncoder {
+	private final CustomEncoder[] encoders;
+	private final int[] offset; // Do not reset encoders, just store an offset value
+	private PIDSourceType pidSource;
+	private boolean reverseDirection;
+	private double distancePerPulse;
+	private final double distanceTolerance;
+	private final double rateTolerance;
+	private static final double DEFAULT_DISTANCE_TOLERANCE = 10;
+	private static final double DEFAULT_RATE_TOLERANCE = 10;
```
leijurv: Protected?

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+		this.distancePerPulse = distancePerPulse;
+		for (CustomEncoder encoder : encoders) {
+			encoder.setDistancePerPulse(distancePerPulse);
+		}
+	}
+	
+	@Override
+	public void reset() {
+		for (int i = 0; i < 2; i++) {
+			offset[i] = encoders[i].get();
```
leijurv: For loop?

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+	
+	@Override
+	public double getDistancePerPulse() {
+		return distancePerPulse;
+	}
+	
+	@Override
+	public void setDistancePerPulse(double distancePerPulse) {
+		this.distancePerPulse = distancePerPulse;
+		for (CustomEncoder encoder : encoders) {
```
leijurv: Enhanced for each loop?

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+	 * This respects the original inversion state of each encoder when constructed,
+	 * and will only invert encoders if this.getReverseDirection() != the input.
+	 *
+	 * @param reverseDirection
+	 *        The state of inversion, true is inverted.
+	 */
+	@Override
+	public void setReverseDirection(boolean reverseDirection) {
+		if (getReverseDirection() != reverseDirection) {
+			for (CustomEncoder encoder : encoders) {
```
leijurv: Enhanced for each loop?

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+	/**
+	 * Sets the direction inversion of all encoder substituents.
+	 * This respects the original inversion state of each encoder when constructed,
+	 * and will only invert encoders if this.getReverseDirection() != the input.
+	 *
+	 * @param reverseDirection
+	 *        The state of inversion, true is inverted.
+	 */
+	@Override
+	public void setReverseDirection(boolean reverseDirection) {
```
leijurv: Unclear naming, local variable hides a field, using this. For differentiation is confusing

custom/sensors/EncoderPair.java
```diff
@@ -0,0 +1,289 @@
...
+		}
+		catch (Exception e) {
+			LogKitten.ex(e);
+			return 0;
+		}
+	}
+	
+	@Override
+	public double getDistanceSafely() throws InvalidSensorException {
+		return ((encoders[0].getDistanceSafely() - offset[0] * encoders[0].getDistancePerPulse()) + (encoders[1].getDistanceSafely() - offset[1] * encoders[1].getDistancePerPulse())) / 2.0;
```
leijurv: No for loop here?

carturn:
Is this something we actually want to do in general?
I would prefer not to waste time cleaning this up if we are not going to use it.

leijurv:
Aren't there various encoder pairs on the chassis?

custom/sensors/EncoderPair.java
carturn: This has generally been our convention for similar setters; I am not sure this function is different enough to warrant a change.

custom/sensors/EncoderPair.java
carturn: Is there an alternative that would work better here?

carturn:
This is partially tested; should we test the PID capabilities of this before merging? I am inclined to think that given that the getDistance and getDifference functions work, the PID will work as well.

PR #146, opened 2017-01-27
Fix Velocity PID set() Implementation

ajnadel:
The current implementation of `SensorMotor`'s `set(double speed)` method used the same value to set the setpoint on the `MotorController` as well as for a fallback raw set() on the motor.

This PR fixes that behavior and gives the following functionality:
- set(double speed) sets velocity PID from -1 to 1
- set(double distancePerSecond) sets velocity PID setpoint to an actual velocity value


subsystems/motor/SensorMotor.java
```diff
@@ -6,108 +6,153 @@
...
+	public void setVelocity(double distancePerSecond) {
+		PIDSourceType previousSensorSourceType = motionController.getSensorSourceType();
+		motionController.setSensorSourceType(PIDSourceType.kRate);
+		motionController.setSetpoint(distancePerSecond);
+		motionController.enable();
+		double speed = motionController.get();
+		motionController.setSensorSourceType(previousSensorSourceType);
+		LogKitten.v(getName() + " set to velocity " + distancePerSecond + " at speed " + speed);
+		super.set(speed);
+	}
```
carturn: This will work, and I think there could be good reasons for doing it, but I would like to note the possibility of having this function try/catch the setVelocitySafely function. setPositionSafely is in a similar state.
Is there a reason for either?

PR #150, opened 2017-02-02
Added ServoSubsystem

c0d3rman:
Added a Subsystem that handles servos to the standard library.

### DIFF
--- a/subsystems/motor/ServoSubsystem.java
+++ b/subsystems/motor/ServoSubsystem.java
@@ -60,11 +60,10 @@ public class ServoSubsystem extends Subsystem {
 	}
 
 	/**
-	 * A class that wraps around a variable number of SpeedController objects to give them Subsystem functionality.
-	 * Can also modify their speed with a SpeedModifier for things like scaling or brownout protection.
+	 * A class that wraps around a variable number of Servo objects to give them Subsystem functionality.
 	 *
-	 * @param motors
-	 *        The SpeedControllers in this subsystem.
+	 * @param servos
+	 *        The Servos in this subsystem.
 	 *        Can be a single Servo or multiple Servo.
 	 */
 	public ServoSubsystem(Servo... servos) {
--- a/subsystems/motor/ServoSubsystem.java
+++ b/subsystems/motor/ServoSubsystem.java
@@ -106,7 +106,7 @@ public class ServoSubsystem extends Subsystem {
 	 * @return The angle in degrees to which the servo is set.
 	 */
 	public double getAngle() {
-		return convertPositionToAngle(get());
+		return ServoSubsystem.convertPositionToAngle(get());
 	}
 
 	/**
@@ -120,7 +120,7 @@ public class ServoSubsystem extends Subsystem {
 	 */
 	public void set(double position) {
 		if (isInverted) {
-			position = invertPosition(lastPosition);
+			position = ServoSubsystem.invertPosition(position);
 		}
 		for (Servo servo : servos) {
 			servo.set(position);
@@ -149,7 +149,7 @@ public class ServoSubsystem extends Subsystem {
 	 */
 	public void setAngle(double degrees) {
 		if (isInverted) {
-			degrees = invertDegree(degrees);
+			degrees = ServoSubsystem.invertDegree(degrees);
 		}
 		for (Servo servo : servos) {
 			servo.setAngle(degrees);
@@ -177,7 +177,7 @@ public class ServoSubsystem extends Subsystem {
 	 */
 	public void setInverted(boolean shouldBeInverted) {
 		if (shouldBeInverted != isInverted) {
-			lastPosition = invertPosition(lastPosition);
+			lastPosition = ServoSubsystem.invertPosition(lastPosition);
 		}
 		isInverted = shouldBeInverted;
 	}
@@ -188,7 +188,7 @@ public class ServoSubsystem extends Subsystem {
 	 * 
 	 * @return the range of the servo in degrees
 	 */
-	protected double getServoAngleRange() {
+	protected static double getServoAngleRange() {
 		return ServoSubsystem.MAX_SERVO_ANGLE - ServoSubsystem.MIN_SERVO_ANGLE;
 	}
 
@@ -199,8 +199,8 @@ public class ServoSubsystem extends Subsystem {
 	 *        servo set value to convert to degrees. Should be in the range [0, 1]
 	 * @return the value converted to degrees
 	 */
-	protected double convertPositionToAngle(double value) {
-		return value * getServoAngleRange() + ServoSubsystem.MIN_SERVO_ANGLE;
+	protected static double convertPositionToAngle(double value) {
+		return value * ServoSubsystem.getServoAngleRange() + ServoSubsystem.MIN_SERVO_ANGLE;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


subsystems/motor/ServoSubsystem.java
```diff
@@ -0,0 +1,242 @@
...
+import org.usfirst.frc4904.standard.commands.Idle;
+import edu.wpi.first.wpilibj.Servo;
+import edu.wpi.first.wpilibj.command.Subsystem;
+
+public class ServoSubsystem extends Subsystem {
+	protected final Servo[] servos;
+	protected boolean isInverted;
+	protected double lastPosition;
+	// Constants from wpilib's Servo.java.
+	private static final double kMaxServoAngle = 180.0;
```
HowardStark: Should be public and dependency injected to allow for more control.

subsystems/motor/ServoSubsystem.java
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	// Internal helper functions
+	/**
+	 * Get the range of the servo in degrees
+	 * 
+	 * @return the range of the servo in degrees
+	 */
+	protected double getServoAngleRange() {
+		return ServoSubsystem.kMaxServoAngle - ServoSubsystem.kMinServoAngle;
```
HowardStark: Should use passed in values

subsystems/motor/ServoSubsystem.java
```diff
@@ -0,0 +1,242 @@
...
+
+	/**
+	 * Convert a servo set value to degrees.
+	 * 
+	 * @param value
+	 *        servo set value to convert to degrees. Should be in the range [0, 1]
+	 * @return the value converted to degrees
+	 */
+	protected double convertPositionToAngle(double value) {
+		return value * getServoAngleRange() + ServoSubsystem.kMinServoAngle;
```
HowardStark: Should use the passed in values

subsystems/motor/ServoSubsystem.java
```diff
@@ -0,0 +1,242 @@
...
+
+	/**
+	 * Convert a degree to a servo set value.
+	 * 
+	 * @param degrees
+	 *        the servo degree to convert to a servo set
+	 * @return a servo set value in the range [0, 1] (as long as the input degree was in the servo's range)
+	 */
+	protected double convertDegreesToValue(double degrees) {
+		return ((degrees - ServoSubsystem.kMinServoAngle)) / getServoAngleRange();
```
HowardStark: Should use the passed in values.

HowardStark:
Aside from the changes I reviewed, it looks good.

subsystems/motor/ServoSubsystem.java
c0d3rman: I don't think so - this is an internal helper function that is copied directly from Servo

HowardStark:
Looks amazing!

subsystems/motor/ServoSubsystem.java
carturn: This could use Util.range, which would be more consistent with the rest of standard.

subsystems/motor/ServoSubsystem.java
```diff
@@ -0,0 +1,242 @@
...
+	 * Set whether this entire servo is inverted.
+	 * Note that this will also convert the values returned by {@link #get() get()}, {@link #getPosition() getPosition()},
+	 * and {@link #getAngle() getAngle()} to the new coordinate system, so anything continuously reading these methods
+	 * will see a discontinuity.
+	 * 
+	 * @param isInverted
+	 *        The state of inversion, true is inverted.
+	 */
+	public void setInverted(boolean isInverted) {
+		if (isInverted != this.isInverted) {
```
leijurv: `^` is more fun than `!=` (I love xor)

leijurv: Also, bad practice to have a local variable with the same name as a field, and differentiating between the two via usage vs non usage of `this.`

leijurv: Perhaps call the argument newInversion idk

subsystems/motor/ServoSubsystem.java
```diff
@@ -0,0 +1,242 @@
...
+
+	/**
+	 * Invert the given value (0 becomes 1, 1 becomes 0)
+	 * 
+	 * @param value
+	 *        The value to invert. Should be in the range [0, 1]
+	 * @return the inverted value
+	 */
+	protected double invertPosition(double value) {
+		return -1 * value + 1;
```
leijurv: So... `return 1 - value;`? Am I missing something here?

subsystems/motor/ServoSubsystem.java
```diff
@@ -0,0 +1,242 @@
...
+
+import org.usfirst.frc4904.standard.commands.Idle;
+import edu.wpi.first.wpilibj.Servo;
+import edu.wpi.first.wpilibj.command.Subsystem;
+
+public class ServoSubsystem extends Subsystem {
+	protected final Servo[] servos;
+	protected boolean isInverted;
+	protected double lastPosition;
+	// Constants from wpilib's Servo.java.
```
leijurv: Is it possible to set `protected static final double kMaxServoAngle = wpilib.Servo.kMaxServoAngle`?
Also why does it start with `k`? Shouldn't final constants have names in all caps separated by underscores?

subsystems/motor/ServoSubsystem.java
```diff
@@ -0,0 +1,242 @@
...
+	 *        This does not override the individual inversions of the servos.
+	 * @param servos
+	 *        The Servos in this subsystem.
+	 *        Can be a single Servo or multiple Servos.
+	 */
+	public ServoSubsystem(String name, boolean isInverted, Servo... servos) {
+		super(name);
+		this.isInverted = false;
+		this.servos = servos;
+		setInverted(isInverted);
```
leijurv: Why does it set `this.isInverted = false;` then call `setInverted`

subsystems/motor/ServoSubsystem.java
c0d3rman: Ahahah I'm stupid

subsystems/motor/ServoSubsystem.java
c0d3rman: Believe it or not I was originally gonna use `^`

subsystems/motor/ServoSubsystem.java
c0d3rman: `wpilib.Servo.kMaxServoAngle` is private (only reason this constant exists)
It starts with `k` because I just copied these from Servo. The name change is a good idea though.

subsystems/motor/ServoSubsystem.java
c0d3rman: whoops

c0d3rman:
@leijurv Please re-review

ajnadel:
Tested on Rasputin!

subsystems/motor/ServoSubsystem.java
```diff
@@ -0,0 +1,241 @@
...
+	 *        Inverts the direction of all of the Servos.
+	 *        This does not override the individual inversions of the servos.
+	 * @param servos
+	 *        The Servos in this subsystem.
+	 *        Can be a single Servo or multiple Servos.
+	 */
+	public ServoSubsystem(String name, boolean isInverted, Servo... servos) {
+		super(name);
+		this.servos = servos;
+		setInverted(isInverted);
```
ajnadel: I agree -- let's just replace this with `this.isInverted = isInverted;`

ajnadel:
Tested on Rasputin!

PR #151, opened 2017-02-03
CapSpeedModifier

NikhilSuresh24:
Added a CapSpeedModifier that implements speedmodifier but sets speeds according to an upper and lower bound given to it


leijurv:
I would suggest using limitValue from Util.Range for consistency (from https://github.com/RoboticsTeam4904/standard/blob/master/Util.java)

c0d3rman:
Agreed, good catch @leijurv

subsystems/motor/speedmodifiers/CapSpeedModifier.java
```diff
@@ -0,0 +1,23 @@
+package org.usfirst.frc4904.standard.subsystems.motor.speedmodifiers;
+
+
+public class CapSpeedModifier implements SpeedModifier {
+	public final double maxSpeed;
+	public final double minSpeed;
+	
+	public CapSpeedModifier(double max, double min) {
+		maxSpeed = max;
```
c0d3rman: By convention we use the same name for constructor arguments that we do for properties. So this line would look like
```
this.maxSpeed = maxSpeed
```

subsystems/motor/speedmodifiers/CapSpeedModifier.java
```diff
@@ -0,0 +1,23 @@
...
+		minSpeed = min;
+	}
+	
+	@Override
+	public double modify(double speed) {
```
c0d3rman: As @leijurv pointed out, although this is a perfectly fine implementation, for consistency we should make a `Util.Range` in the constructor, and then call `limitValue` on it here.

ajnadel:
@NikhilSuresh24 (@c0d3rman) maybe consider adding a constructor that takes in an existing Util.Range in addition to that one?

subsystems/motor/speedmodifiers/CapSpeedModifier.java
```diff
@@ -0,0 +1,21 @@
+package org.usfirst.frc4904.standard.subsystems.motor.speedmodifiers;
+
+
+import org.usfirst.frc4904.standard.Util;
+
+public class CapSpeedModifier implements SpeedModifier {
+	public Util.Range speedCap;
+	
+	public CapSpeedModifier(double minSpeed, double maxSpeed) {
+		speedCap = new Util.Range(minSpeed, maxSpeed);
```
leijurv: Perhaps use a self constructor (e.g. replace with `this(new Util.Range(minSpeed, maxSpeed));`)

subsystems/motor/speedmodifiers/CapSpeedModifier.java
```diff
@@ -0,0 +1,21 @@
+package org.usfirst.frc4904.standard.subsystems.motor.speedmodifiers;
+
+
+import org.usfirst.frc4904.standard.Util;
+
+public class CapSpeedModifier implements SpeedModifier {
+	public Util.Range speedCap;
```
leijurv: Also please make the speedCap either protected or private. If it needs to be accessed, add a getter

c0d3rman:
@leijurv Made the changes you requested

leijurv:
Noticed something while taking another glance, I'm not sure if this is the convention, but you can do `import org.usfirst.frc4904.standard.Util.Range;` and then refer to the class as just `Range` and not  `Util.Range`.

Now it feels like importing `import org.usfirst.frc4904;` and using `standard.Util.Range` everywhere (even though you can't actually do _that_ lol)

c0d3rman:
Since `Range` is a class within `Util` I think this is proper usage

leijurv:
"proper" says who =P

c0d3rman:
Says me

leijurv:
GOTEM

ajnadel:
Leif takes the L

leijurv:
ill take your L

PR #152, opened 2017-02-04
Better CAN logging

carturn:
Readded CAN ID to CANSensor error message.

Nobody names their CANSensors, so this helps determine which node is having issues.


carturn:
To the surprise of everyone, this code actually works.

PR #153, opened 2017-02-08
Removed CAN loop

carturn:
Removed the 5ms loop checking for CAN messages.
Now checks once and then returns null.
This is equally reliable, with less code.


carturn:
Still works after master merge.

custom/CustomCAN.java
```diff
@@ -60,14 +59,10 @@ protected ByteBuffer readBuffer() {
...
-		while (System.currentTimeMillis() - start < CustomCAN.CAN_MAX_READ_WAIT) {
-			try {
-				response = CANJNI.FRCNetCommCANSessionMuxReceiveMessage(idBuffer, 0x1fffffff, timestamp);
-				break;
-			}
-			catch (CANMessageNotFoundException e) {}
+		try {
+			response = CANJNI.FRCNetCommCANSessionMuxReceiveMessage(idBuffer, 0x1fffffff, timestamp);
 		}
+		catch (CANMessageNotFoundException e) {}
```
c0d3rman: Should we maybe print something here?

c0d3rman: @carturn

carturn: I would agree with you completely if I knew when that exception happens and what it means.
(CAN message not found does seem somewhat self-explanatory, but I am concerned that things are not what they seem behind the JNI wall).
It would not hurt to add a warning for it I suppose.

PR #154, opened 2017-02-09
MotionController Override

ave4224:
Allow MotionControllers to be overridden with an instance method. When overridden, they only do cursory calculations, like feed-forward multiplication of the input.


custom/motioncontrollers/MotionController.java
```diff
@@ -256,27 +261,57 @@ public void enable() {
...
+	 */
+	public void kill() {
+		disable();
+		alive = false;
+	}
+	
+	/**
+	 * Resuscitates a controller, allowing disable() and enable() to be called again.
+	 */
+	public void resuscitate() {
```
carturn: Is there a shorter/easier to spell verb?

subsystems/motor/SensorMotor.java
```diff
@@ -11,59 +11,76 @@
 public abstract class SensorMotor extends Motor {
 	protected final MotionController motionController;
 	private boolean isMotionControlEnabled;
-
-	public SensorMotor(String name, boolean inverted, SpeedModifier speedModifier, MotionController motionController,
-		SpeedController... motors) {
+	
+	public SensorMotor(String name, boolean inverted, SpeedModifier speedModifier, MotionController motionController, SpeedController... motors) {
```
carturn: Please update your formatter before merging.
The split in this line is correct under the new paradigm of _avoid sideways scrolling_

custom/motioncontrollers/MotionController.java
c0d3rman: perhaps have a single `setOverride` method? or an `enableOverride` and `disableOverride`?

ajnadel:
Of note, I changed the essence of this PR. Now, it doesn't disallow disable() and enable(), but rather it only multiplies F*input. If we disallowed enabling, we'd have to simultaneously start using set() on the motors directly, as all of the setSetpoint() methods would not work because disabling() [which happens once at the time of override] would cancel the MotionControllerTask and the changing setpoint would never be reflected on the motor.

custom/motioncontrollers/MotionController.java
```diff
@@ -278,6 +286,47 @@ public boolean isEnabled() {
...
 
 	/**
+	 * Set whether or not the motion controller
+	 * is overridden.
+	 * 
+	 * @see #startOverriding()
+	 * @see #stopOverriding()
+	 */
+	public void setOverride(boolean override) {
+		overridden = override;
```
c0d3rman: Please use the convention `this.overridden = overridden`

PR #157, opened 2017-02-12
Propagating CANMessageNoFoundException

carturn:
Added the warning @c0d3rman suggested in #153.


custom/CustomCAN.java
```diff
@@ -62,7 +63,9 @@ protected ByteBuffer readBuffer() {
 		try {
 			response = CANJNI.FRCNetCommCANSessionMuxReceiveMessage(idBuffer, 0x1fffffff, timestamp);
 		}
-		catch (CANMessageNotFoundException e) {}
+		catch (CANMessageNotFoundException e) {
+			LogKitten.w(e.getMessage()); // We are not actually sure when this error is used (no docs), so just warn for now
```
c0d3rman: Are you sure `LogKitten.e` isn't appropriate? If we don't know what the error is, all the more reason to err on the side of caution (we can't guarantee it's safe).

carturn: Actually, I think the way I am doing this now is entirely wrong. Without the loop, there is no reason not to pass on the error.

carturn:
OK, instead of just logging, now passing the CANMessageUnavailableException up. This is similar to the CANMessageNotFoundException, but instead of being a RuntimeException (which can occur unexpectedly), it is an Exception so that we know when it needs to be caught.

carturn:
Still tested.

PR #158, opened 2017-02-13
Add EnableableModifier to standard

ajnadel:
Improve enable-able modifier and move it to `standard` from `2017-Code`.

### DIFF
--- a/commands/motor/speedmodifiers/SetEnableableModifier.java
+++ b/commands/motor/speedmodifiers/SetEnableableModifier.java
@@ -21,6 +21,6 @@ public class SetEnableableModifier extends Command {
 
 	@Override
 	protected boolean isFinished() {
-		return false;
+		return true;
 	}
 }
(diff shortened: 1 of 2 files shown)


ajnadel:
While this is @c0d3rman's change, I've verbally approved it.

commands/motor/speedmodifiers/SetEnableableModifier.java
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	@Override
+	protected void initialize() {
+		modifier.setEnabled(enable);
+	}
+
+	@Override
+	protected boolean isFinished() {
+		return false;
```
ajnadel: Should this return true?
Or else disable the modifier when it ends.

nacgarg: Returning false seems fine, you could also return (enable == modifier.isEnabled()) to make it clearer.

PR #159, opened 2017-02-13
Added another CANEncoder constructor

c0d3rman:
Added a constructor for just a name and an ID (could be used in `2017-Code`'s `RobotMap`).


c0d3rman:
It's true, but for quick hax I think we should keep the non-named constructors - if someone's trying to debug and having trouble they can just add names. (Especially now that all permutations of the constructor are available)

PR #160, opened 2017-02-13
Added convenience Until/While commands

c0d3rman:
Added convenience commands using Boolean Suppliers to run something/wait while/until something is true. Could be used in `2017-Code`'s `Shoot`.


c0d3rman:
I wasn't sure if these should inherit each other with `Noop`s and `Supplier<Boolean>`s that invert `Supplier<Boolean>`s.

ajnadel:
@c0d3rman It's probably fine as is.

PR #161, opened 2017-02-15
PDP class reading CAN directly

carturn:
The PowerDistributionPanel class throws exceptions that can not be caught (based on tests by @c0d3rman). In order to get around this, the DriverStation voltage can be read, but this potentially has latency and does not include amperage data.
This pull request basically reads the CAN data directly, which allows us to control the exceptions.


carturn:
Seems to return values very close to those returned by the PowerDistributionPanel class, which is expected considering that the PDP only sends data every 50-100ms, which is ample time for things to change.

carturn:
OK, now it actually works. Side note, we need to slow our ramping, the Napoleon's flywheel drew 40 ish amps while spinning up and only 16 while idle.

PR #162, opened 2017-02-16
Added null check

carturn:
Added null check for fallbackCommand to MotorPositionConstant.


carturn:
Still tested.

carturn:
Still tested.

PR #163, opened 2017-02-17
Switched to getting voltage from the DS

c0d3rman:
To avoid uncatchable catastrophic CTRE CAN Receive Timeout. Robot now still operates VictorSPs, pneumatics, and servos when CAN bus is cut.

### DIFF
--- a/subsystems/motor/speedmodifiers/AccelerationCap.java
+++ b/subsystems/motor/speedmodifiers/AccelerationCap.java
@@ -3,6 +3,7 @@ package org.usfirst.frc4904.standard.subsystems.motor.speedmodifiers;
 
 import org.usfirst.frc4904.standard.LogKitten;
 import org.usfirst.frc4904.standard.custom.sensors.PDP;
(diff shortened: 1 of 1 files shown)


subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -61,10 +62,11 @@ public AccelerationCap(PDP pdp) {
 	 */
 	@Override
 	public double modify(double inputSpeed) {
+		double voltage = DriverStation.getInstance().getBatteryVoltage(); // TODO add default for when data isn't available
```
ajnadel: WPILib isn't transparent about what errors could be thrown -- we (I) should try to trigger errors and see what happens.

carturn:
We should use #161 instead, lower latency and includes current.

subsystems/motor/speedmodifiers/AccelerationCap.java
ajnadel: We'll figure it out along the way.

PR #164, opened 2017-02-18
Added default constructor to PDP

carturn:


c0d3rman:
Bypassing testing because it's build season and I need this merged.

PR #167, opened 2017-02-18
Made MotorConstant nameable

c0d3rman:


c0d3rman:
The reason I did this is because I needed it for debugging on Alexandra

PR #168, opened 2017-02-18
Better Brownout Management

carturn:
Improved the brownout prevention system (hopefully).
This should allow the system to get closer to brownouts while maintaining protection from them.
In addition, the ramping system was made truly linear, so it should work better now.


carturn:
This is tested, but the soft threshold system is fairly hacky. If someone has a better idea of how to estimate future voltage drop given present voltage drop for all motors, that would be helpful.

carturn:
The ramping portion of this is stable. The brownout portion seems to work, but needs more testing with more dead batteries.

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -41,15 +52,69 @@ public AccelerationCap(PDP pdp, double softStopVoltage, double hardStopVoltage)
...
-		this(pdp, 11.0, 10.0);
+	public AccelerationCap(PDP pdp, int... channels) {
+		this(pdp, AccelerationCap.DEFAULT_HARD_STOP_VOLTAGE, channels);
+	}
+
+	protected double calculate(double inputSpeed) {
+		double deltaTime = (System.currentTimeMillis() - lastUpdate) / 1000.0;
+		lastUpdate = System.currentTimeMillis();
+		if (deltaTime > AccelerationCap.TIMEOUT_SECONDS) {
+			return 0;
```
c0d3rman: Won't this kill all AccelerationCapped motors if the CAN bus is cut? (which we wanted to avoid)

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -41,15 +52,69 @@ public AccelerationCap(PDP pdp, double softStopVoltage, double hardStopVoltage)
...
+			if (inputSpeed > currentSpeed) {
+				return currentSpeed + AccelerationCap.MAXIMUM_MOTOR_DECREASE_PER_SECOND * deltaTime;
+			}
+			if (inputSpeed < currentSpeed) {
+				return currentSpeed - AccelerationCap.MAXIMUM_MOTOR_DECREASE_PER_SECOND * deltaTime;
+			}
+		}
+		double rampedSpeed = inputSpeed;
+		// Ramping
+		if (Math.abs(currentSpeed - inputSpeed) > AccelerationCap.MAXIMUM_MOTOR_INCREASE_PER_SECOND * deltaTime) {
```
c0d3rman: This is to protect the gearboxes?

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -41,15 +52,69 @@ public AccelerationCap(PDP pdp, double softStopVoltage, double hardStopVoltage)
...
+			if (inputSpeed > currentSpeed) {
+				rampedSpeed = currentSpeed + AccelerationCap.MAXIMUM_MOTOR_INCREASE_PER_SECOND * deltaTime;
+			} else if (inputSpeed < currentSpeed) {
+				rampedSpeed = currentSpeed - AccelerationCap.MAXIMUM_MOTOR_INCREASE_PER_SECOND * deltaTime;
+			}
+		}
+		// After ramping, apply brown-out protection
+		double currentVoltage = pdp.getVoltage(); // Allow fallback to DS voltage
+		if (currentVoltage < hardStopVoltage) { // If we are below hardStopVoltage, start backing off
```
c0d3rman: If hardStopVoltage is 7 and we start backing off then, isn't this too little too late?

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -41,15 +52,69 @@ public AccelerationCap(PDP pdp, double softStopVoltage, double hardStopVoltage)
...
+			rampedSpeed = currentSpeed
+				- AccelerationCap.ANTI_BROWNOUT_BACKOFF_PER_SECOND * Math.signum(currentSpeed) * deltaTime;
+			if (Math.abs(rampedSpeed) <= AccelerationCap.ANTI_BROWNOUT_BACKOFF_PER_SECOND) {
+				return 0;
```
c0d3rman: What is this?

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -41,15 +52,69 @@ public AccelerationCap(PDP pdp, double softStopVoltage, double hardStopVoltage)
...
+			}
+			return rampedSpeed;
+		}
+		// Even if we are still above the hard stop voltage, try to avoid going below next tick
+		try {
+			double currentCurrent = 0.0;
+			for (int channel : channels) {
+				currentCurrent += pdp.getCurrent(channel);
```
c0d3rman: Don't we have a total current function

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -41,15 +52,69 @@ public AccelerationCap(PDP pdp, double softStopVoltage, double hardStopVoltage)
...
+			}
+			double voltageDrop = currentCurrent * pdp.getBatteryResistanceSafely();
+			double deltaVoltageDrop = voltageDrop - lastVoltageDrop;
+			if (currentVoltage < hardStopVoltage + voltageDrop + deltaVoltageDrop * AccelerationCap.TICKS_PER_PDP_DATA) {
+				return currentSpeed;
```
c0d3rman: So we maintain current speed if just above hardStopVoltage but ramp down if below?

subsystems/motor/speedmodifiers/AccelerationCap.java
carturn: No, because lastUpdate is only determined when this function is run, which is always. The idea here was to catch motors being left on and then disabled, which stops calling this function, which can cause a 1 cycle twitch.

subsystems/motor/speedmodifiers/AccelerationCap.java
carturn: Mostly. It should not affect power usage much.

subsystems/motor/speedmodifiers/AccelerationCap.java
carturn: That is fair.

subsystems/motor/speedmodifiers/AccelerationCap.java
carturn: Slowing down in discrete steps, but avoiding overshoot. (e.g. 3.2 - 6.4 = -3.2, which is speeding up again).

subsystems/motor/speedmodifiers/AccelerationCap.java
c0d3rman: Ah I see.

custom/sensors/PDP.java
```diff
@@ -56,36 +56,54 @@ public PDP() {
...
-			cachedChannelCurrents[5] = (((rawArray[6] & 0x3F) << 4) | rawArray[7] & 0xF0) * 0.125;
+			double[] tempCurrents = new double[6];
+			tempCurrents[0] = ((rawArray[0] & 0xFF) << 2 | ((rawArray[1] & 0xC0) >> 6)) * 0.125;
+			tempCurrents[1] = (((rawArray[1] & 0x3F) << 4) | ((rawArray[2] & 0xF0) >> 4)) * 0.125;
+			tempCurrents[2] = (((rawArray[2] & 0x0F) << 6) | ((rawArray[3] & 0x3F) >> 2)) * 0.125;
+			tempCurrents[3] = (((rawArray[3] & 0xC0) << 8) | ((rawArray[4] & 0xFF))) * 0.125;
+			tempCurrents[4] = (((rawArray[5] & 0xFF) << 2) | ((rawArray[6] & 0xC0) >> 6)) * 0.125;
+			tempCurrents[5] = (((rawArray[6] & 0x3F) << 4) | ((rawArray[7] & 0xF0) >> 4)) * 0.125;
+			for (int i = 0; i < 6; i++) {
+				if (tempCurrents[i] < 128) { // deals with occasional issue with PDP reporting 1000+ amps (this is not a bug in this code, it was observed in PowerDistributionPanel as well
```
c0d3rman: Perhaps now that there is actual logic here the three readStatus methods should be generalized into one?

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -42,8 +45,9 @@ public AccelerationCap(PDP pdp, double hardStopVoltage, int... channels) {
 		this.pdp = pdp;
 		this.hardStopVoltage = hardStopVoltage;
 		this.channels = channels;
+		disableCurrent = pdp.getTotalCurrent() > AccelerationCap.REASONABLE_RESTING_CURRENT;
```
c0d3rman: Perhaps we should print a warning if we disable the current?

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -128,7 +143,7 @@ protected double calculate(double inputSpeed) {
 	@Override
 	public double modify(double inputSpeed) {
 		currentSpeed = calculate(inputSpeed);
-		LogKitten.d("AccelerationCap outputed: " + currentSpeed);
+		LogKitten.w("AccelerationCap outputed: " + currentSpeed);
```
c0d3rman: Why is this a warn? Also spelling.

custom/sensors/PDP.java
carturn: I combined 1 & 2, but 3 also has voltage and resistance.

custom/sensors/PDP.java
```diff
@@ -36,17 +38,66 @@ public PDP(int ID) {
...
+		} else if (status == 3) {
+			readStatus3();
+		}
+		if (rawArray != null) {
+			double[] tempCurrents = new double[6];
+			tempCurrents[0] = ((rawArray[0] & 0xFF) << 2 | ((rawArray[1] & 0xC0) >> 6)) * 0.125;
+			tempCurrents[1] = (((rawArray[1] & 0x3F) << 4) | ((rawArray[2] & 0xF0) >> 4)) * 0.125;
+			tempCurrents[2] = (((rawArray[2] & 0x0F) << 6) | ((rawArray[3] & 0x3F) >> 2)) * 0.125;
+			tempCurrents[3] = (((rawArray[3] & 0xC0) << 8) | ((rawArray[4] & 0xFF))) * 0.125;
+			tempCurrents[4] = (((rawArray[5] & 0xFF) << 2) | ((rawArray[6] & 0xC0) >> 6)) * 0.125;
```
c0d3rman: Does this intentionally skip over 4?

custom/sensors/PDP.java
```diff
@@ -36,17 +38,66 @@ public PDP(int ID) {
...
+			for (int i = 0; i < 6; i++) {
+				if (tempCurrents[i] < 128) { // deals with occasional issue with PDP reporting 1000+ amps (this is not a bug in this code, it was observed in PowerDistributionPanel as well
+					cachedChannelCurrents[i + (status - 1) * 6] = tempCurrents[i];
+				}
+			}
+			lastRead = System.currentTimeMillis();
+		}
+	}
+
+	private void readStatus3() { // Status 3 broken out for voltage and resistance
```
c0d3rman: It's kinda weird to have a `readStatus3` but no `readStatus1` or `readStatus2`. Also some logic is reimplemented here. Perhaps make a generalized function reading n values and then call that in `readStatus1` and `readStatus2` with n=6 and then call it in `readStatus3` with n=4 and additional logic?

custom/sensors/PDP.java
```diff
@@ -36,17 +38,66 @@ public PDP(int ID) {
...
+			tempCurrents[1] = (((rawArray[1] & 0x3F) << 4) | ((rawArray[2] & 0xF0) >> 4)) * 0.125;
+			tempCurrents[2] = (((rawArray[2] & 0x0F) << 6) | ((rawArray[3] & 0x3F) >> 2)) * 0.125;
+			tempCurrents[3] = (((rawArray[3] & 0xC0) << 8) | ((rawArray[4] & 0xFF))) * 0.125;
+			for (int i = 0; i < 4; i++) {
+				if (tempCurrents[i] < 128) {
+					cachedChannelCurrents[i + 12] = tempCurrents[i];
+				}
+			}
+			cachedVoltage = (rawArray[6] & 0xFF) * 0.05 + 4.0;
+			cachedResistance = (rawArray[5] & 0xFF) / 1000.0; // in milliOhms
```
c0d3rman: Pedantic but perhaps the order of these should be flipped? (So it's ordered by the index)

custom/sensors/PDP.java
```diff
@@ -75,38 +126,89 @@ public double getVoltage() {
...
 	 */
 	public double getVoltageSafely() throws InvalidSensorException {
-		byte[] rawArray = status3.read();
-		if (rawArray != null) {
-			int rawVoltage = rawArray[6] & 0xFF; // busVoltage is the 7th byte of the third status
-			cachedVoltage = (rawVoltage) * 0.05 + 4.0;
-			lastRead = System.currentTimeMillis();
-		}
+		readStatus3();
 		if (System.currentTimeMillis() - lastRead > PDP.MAX_AGE) {
```
c0d3rman: Since this code is duplicated, maybe break it out into its own function or implement it inside the `readStatus` methods?

custom/sensors/PDP.java
```diff
@@ -75,38 +126,89 @@ public double getVoltage() {
...
+		if (channel < 0) {
+			return 0.0;
+		} else if (channel <= 5) {
+			readStatus(1);
+		} else if (channel <= 11) {
+			readStatus(2);
+		} else if (channel <= 15) {
+			readStatus(3);
+		} else {
+			return 0.0;
```
c0d3rman: We should probably at least log a warning here.

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -41,15 +58,73 @@ public AccelerationCap(PDP pdp, double softStopVoltage, double hardStopVoltage)
...
+		if (!disableCurrent) {
+			newVoltageDrop = voltageDrop;
+			try {
+				newVoltageDrop = pdp.getTotalCurrentSafely() * pdp.getBatteryResistanceSafely();
+			}
+			catch (InvalidSensorException e) {
+				LogKitten.ex(e);
+			}
+			lastVoltageDrop = voltageDrop;
+			if (!new Util.Range(lastVoltageDrop - 0.01, lastVoltageDrop + 0.01).contains(newVoltageDrop)) {
```
c0d3rman: Maybe make the 0.01 a constant?

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -41,15 +58,73 @@ public AccelerationCap(PDP pdp, double softStopVoltage, double hardStopVoltage)
...
+	public AccelerationCap(PDP pdp, int... channels) {
+		this(pdp, AccelerationCap.DEFAULT_HARD_STOP_VOLTAGE, channels);
+	}
+
+	protected double calculate(double inputSpeed) {
+		double deltaTime = (System.currentTimeMillis() - lastUpdate) / 1000.0;
+		lastUpdate = System.currentTimeMillis();
+		// Update current data
+		double newVoltageDrop = 0.0;
+		double lastVoltageDrop = 0.0;
```
c0d3rman: Why are there three variables? (newVoltageDrop, voltageDrop, and lastVoltageDrop)

custom/sensors/PDP.java
```diff
@@ -36,17 +38,66 @@ public PDP(int ID) {
...
+			}
+			cachedVoltage = (rawArray[6] & 0xFF) * 0.05 + 4.0;
+			cachedResistance = (rawArray[5] & 0xFF) / 1000.0; // in milliOhms
+			lastRead = System.currentTimeMillis();
+		}
+	}
+
 	/**
 	 * Gets the current voltage. This is the same for all channels.
 	 * This function defaults to the Driver Station voltage if the PDP becomes disconnected.
```
c0d3rman: This isn't actually true - it just errors. (There are a bunch of comments claiming this, including one in AccelerationCap.)

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -29,10 +40,16 @@
 	 * @param hardStopVoltage
 	 *        Voltage to begin decreasing motor speed at.
 	 */
-	public AccelerationCap(PDP pdp, double softStopVoltage, double hardStopVoltage) {
+	public AccelerationCap(PDP pdp, double hardStopVoltage, int... channels) {
 		this.pdp = pdp;
-		this.softStopVoltage = softStopVoltage;
 		this.hardStopVoltage = hardStopVoltage;
+		this.channels = channels;
+		disableCurrent = pdp.getTotalCurrent() > AccelerationCap.REASONABLE_RESTING_CURRENT;
```
c0d3rman: Is it OK that we only ever check this on construction?

subsystems/motor/speedmodifiers/AccelerationCap.java
```diff
@@ -61,28 +136,8 @@ public AccelerationCap(PDP pdp) {
...
-				outputSpeed = inputSpeed;
-			}
-		} else {
-			outputSpeed = inputSpeed;
-		}
-		LogKitten.d("AccelerationCap input: " + inputSpeed + "\t" + "AccelerationCap output: " + outputSpeed);
-		lastUpdate = System.currentTimeMillis();
-		currentSpeed = outputSpeed;
-		return outputSpeed;
+		currentSpeed = calculate(inputSpeed);
```
c0d3rman: Why was all the logic moved to a separate function?

custom/sensors/PDP.java
```diff
@@ -53,8 +53,15 @@ public PDP() {
...
 	}
 
-	private void readStatus1() {
-		byte[] rawArray = status1.read();
+	private void readStatus(int status) {
+		byte[] rawArray = null;
+		if (status == 1) {
+			rawArray = status1.read();
+		} else if (status == 2) {
+			rawArray = status1.read();
```
carturn: Yes, fixed.

custom/sensors/PDP.java
carturn: Are you referring to rawArray[x] in the first column? If so, yes.

custom/sensors/PDP.java
carturn: It will display an error, but it should return a valid voltage. What is the exception being thrown?

custom/sensors/PDP.java
carturn: I think throwing some invalid input exception would be most appropriate, but that could cause issues. Technically the current on a channel above 15 will be 0.

subsystems/motor/speedmodifiers/AccelerationCap.java
carturn: I am fairly confident that a PDP issue will not appear in the middle of a match, and this is one of the best times that we have a guarantee that all motor controllers are disabled.

custom/sensors/PDP.java
c0d3rman: Yes, I was. Out of curiousity, why?

custom/sensors/PDP.java
c0d3rman:
```
throw new InvalidSensorException("Can not read voltage from PDP");
```
If the PDP becomes disconnected it just throws an `InvalidSensorException` and doesn't return anything.

custom/sensors/PDP.java
c0d3rman: Agreed. What I meant was that whoever tried to access a port above 15 probably didn't mean to, so a `LogKitten.w` might prevent some annoyance in the future.

custom/sensors/PDP.java
c0d3rman: (This is still relevant despite being on outdated code)

subsystems/motor/speedmodifiers/AccelerationCap.java
c0d3rman: OK

subsystems/motor/speedmodifiers/AccelerationCap.java
carturn: This allows for multiple exit points from the calculation while still storing currentSpeed.

c0d3rman: OK

custom/sensors/PDP.java
carturn: That should be caught in this function. Strange.

PR #169, opened 2017-02-18
Teensy joystick

NikhilSuresh24:
made teensy controller with 30 custombuttons

### DIFF
--- a/custom/controllers/TeensyController.java
+++ b/custom/controllers/TeensyController.java
@@ -5,13 +5,15 @@ import edu.wpi.first.wpilibj.DriverStation;
 import edu.wpi.first.wpilibj.Joystick;
 
 public class TeensyController extends Joystick implements Controller {
-	public CustomButton[] buttons = new CustomButton[30];
+	public int numButtons;
+	public CustomButton[] buttons = new CustomButton[numButtons];
 	public final int port;
 
-	public TeensyController(int port) {
+	public TeensyController(int port, int numButtons) {
 		super(port);
 		this.port = port;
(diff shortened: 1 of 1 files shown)


custom/controllers/TeensyController.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class TeensyController extends Joystick implements Controller {
+	public CustomButton[] buttons = new CustomButton[30];
+	public final int port;
+
+	public TeensyController(int port) {
+		super(port);
+		this.port = port;
+		for (int index = 1; index < 31; index++) {
+			buttons[index] = new CustomButton(this, index);
```
carturn: This is going to cause an error, the maximum index of the array if 29.

custom/controllers/TeensyController.java
```diff
@@ -0,0 +1,34 @@
+package org.usfirst.frc4904.standard.custom.controllers;
+
+
+import edu.wpi.first.wpilibj.DriverStation;
+import edu.wpi.first.wpilibj.Joystick;
+
+public class TeensyController extends Joystick implements Controller {
+	public CustomButton[] buttons = new CustomButton[30];
```
carturn: Why not make the configurable?

custom/controllers/TeensyController.java
carturn: I would suggest making the buttons zero-indexed, the one-indexing on the other joysticks is suboptimal.

carturn:
Also, please test before merging.

PR #172, opened 2017-02-19
Fix NavX Rate for PID

ajnadel:
We currently have code that will stop the NavX from returning extraneous values. This fixes a potential bug and differentiates this logic between the get() and getRate() methods.


custom/sensors/NavX.java
```diff
@@ -25,13 +27,17 @@ public NavX(SerialPort.Port port) {
...
+		} else {
+			return getYaw();
+		}
 	}
 
 	@Override
 	public double getRate() {
 		double rate = super.getRate();
-		if (Math.abs(rate) > Math.abs(lastYawRate) + MAX_DEGREES_PER_TICK) {
+		if (Math.abs(rate / lastYawRate) > NavX.MAX_RATE_INCREASE_FACTOR_PER_TICK) {
```
carturn: I think this might cause issues going from zero to non-zero values.

custom/sensors/NavX.java
```diff
@@ -61,7 +67,7 @@ public float getRawYaw() {
 	@Override
 	public float getPitch() {
 		float pitch = super.getPitch();
-		if (Math.abs(pitch) > Math.abs(lastPitch) + MAX_DEGREES_PER_TICK) {
+		if (Math.abs(pitch) > Math.abs(lastPitch) + NavX.MAX_DEGREES_PER_TICK) {
```
carturn: Why not change this one to `Math.abs(pitch - lastPitch)` as well?

carturn:
I am generally concerned about runtime exceptions, but this is better than the previous runtime exceptions.

PR #173, opened 2017-02-19
Made CANTalonEncoder update faster

carturn:
We used to be using the getEnc* functions, which based on my reading of https://github.com/wpilibsuite/allwpilib/blob/55346e28d02b043770c76f7ec9f2aff82f845ceb/hal/lib/athena/ctre/CanTalonSRX.cpp only get updated every 100ms.
This pull request switches to getPosition and getSpeed, which update every 20ms (should make PID loops much better).


custom/sensors/CANTalonEncoder.java
```diff
@@ -21,129 +23,129 @@ public CANTalonEncoder(String name, CANTalon talon, boolean reverseDirection, do
...
 	}
-	
+
 	@Override
 	public double getRate() {
 		if (reverseDirection) {
-			return talon.getEncVelocity() * -10.0 * distancePerPulse;
-			// getEncVelocity must be converted from ticks to 100ms to ticks per second, so *10.
+			return talon.getSpeed() * -10.0 * distancePerPulse;
+			// getSpeed must be converted from ticks to 100ms to ticks per second, so *10.
```
c0d3rman: Is this still correct now that we read once per 20ms?

c0d3rman: Also, perhaps we should keep track of the actual time that's passed?

carturn: Based on code from here: http://first.wpi.edu/FRC/roborio/release/docs/cpp/classCANTalon.html#a36e486b58bc5ce1b218d36468900bb1b, `The speed units will be in the sensor's native ticks per 100ms.`, so that is still correct.
Also, the derivative is taken on the Talon (similar to how the Teensys do it), so it is more precise and we do not have to care.

c0d3rman: OK

PR #175, opened 2017-02-21
Solenoid shifters last shift

NikhilSuresh24:
autoSolenoidShifters(necessary for autoshifting and disable-auto-shifting)


subsystems/chassis/AutoSolenoidShifters.java
```diff
@@ -0,0 +1,64 @@
...
+		lastManualShift = System.currentTimeMillis();
+	}
+
+	@Override
+	public void shift(ShiftState state) {
+		super.shift(state);
+		lastManualShift = System.currentTimeMillis();
+	}
+
+	public void shift(ShiftState state, boolean isAutoShift) {
```
c0d3rman: Make this a separate method, not just taking in a boolean.

subsystems/chassis/AutoSolenoidShifters.java
```diff
@@ -0,0 +1,64 @@
...
+		if (isAutoShift) {
+			lastAutoShift = System.currentTimeMillis();
+			switch (state) {
```
c0d3rman: Why not just call `super.shift(state)`?

subsystems/chassis/AutoSolenoidShifters.java
c0d3rman: Also, what about a toggle shift method?

c0d3rman:
This is jank enough so that we're going to keep it in yearly code pending a better system.

PR #177, opened 2017-02-25
Added exception handling to CustomCAN

c0d3rman:
Added exception handling so that the CAN bus being cut doesn't kill the bot.
The stacktrace printing logic should eventually be moved to LogKitten.

### DIFF
--- a/custom/CustomCAN.java
+++ b/custom/CustomCAN.java
@@ -1,8 +1,6 @@
 package org.usfirst.frc4904.standard.custom;
 
 
-import java.io.PrintWriter;
-import java.io.StringWriter;
 import java.nio.ByteBuffer;
 import java.nio.IntBuffer;
 import org.usfirst.frc4904.standard.LogKitten;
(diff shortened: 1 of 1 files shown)


custom/CustomCAN.java
```diff
@@ -39,9 +42,26 @@ public String getName() {
...
-	 * @throws IllegalArgumentException
 	 */
 	public void write(byte[] data) {
+		try {
+			writeSafely(data);
+		}
+		catch (Exception e) {
+			StringWriter sw = new StringWriter();
+			e.printStackTrace(new PrintWriter(sw));
+			LogKitten.e(sw.toString());
```
carturn: Why not just `LogKitten.ex(e)`?

custom/CustomCAN.java
```diff
@@ -39,9 +42,26 @@ public String getName() {
...
+		}
+	}
+
+	/**
+	 * Used to write data to the device.
+	 *
+	 * @param data
+	 *        Data to be written. Should be EXACTLY 8 bytes long ONLY.
+	 * @throws IllegalArgumentException
```
carturn: Is the error it throws IllegalArgumentException or something CAN related?

custom/CustomCAN.java
carturn: Also, catching specific errors is better.

PR #180, opened 2017-03-04
Threshold Command

billwpierce:
I Took the threshold command and added it to standard. Will run the command when it gets past a certain point. Once merged, we should add to 2017-Code.

### DIFF
--- a/commands/ThresholdCommand.java
+++ b/commands/ThresholdCommand.java
@@ -27,7 +27,7 @@ public class ThresholdCommand<T extends Comparable<T>> extends Command {
 	}
 
 	protected boolean pastThreshold() {
-		return ((supplier.get().compareTo(threshold) >= 0) ^ invert);
+		return ((supplier.get().compareTo(threshold) >= 0) != invert);
 	}
 
 	@Override
--- a/commands/ThresholdCommand.java
+++ b/commands/ThresholdCommand.java
@@ -1,0 +1,44 @@
+package org.usfirst.frc4904.standard.commands;
+
+
+import java.util.function.Supplier;
+import edu.wpi.first.wpilibj.command.Command;
+
+/**
+ *
+ */
+public class ThresholdCommand extends Command {
+	protected final Command command;
+	protected final Supplier<Double> axis;
+	protected final double threshold;
+	protected final boolean invert;
+
+	public ThresholdCommand(Command command, Supplier<Double> axis, double threshold, boolean invert) {
+		this.command = command;
+		this.axis = axis;
+		this.threshold = threshold;
+		this.invert = invert;
+	}
+
+	public ThresholdCommand(Command command, Supplier<Double> axis, double threshold) {
+		this(command, axis, threshold, false);
+	}
+
+	protected boolean pastThreshold(double value) {
+		return (!invert && value >= threshold) || (invert && value <= threshold);
+	}
+
+	@Override
+	protected void execute() {
+		if (pastThreshold(axis.get()) && !command.isRunning()) {
+			command.start();
+		} else if (!pastThreshold(axis.get()) && command.isRunning()) {
+			command.cancel();
+		}
+	}
+
+	@Override
+	protected boolean isFinished() {
+		return false;
+	}
+}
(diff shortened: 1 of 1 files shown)


commands/ThresholdCommand.java
```diff
@@ -0,0 +1,44 @@
+package org.usfirst.frc4904.standard.commands;
+
+
+import java.util.function.Supplier;
+import edu.wpi.first.wpilibj.command.Command;
+
+/**
+ *
```
c0d3rman: We should add a javadoc (or remove this empty one).

commands/ThresholdCommand.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class ThresholdCommand extends Command {
+	protected final Command command;
+	protected final Supplier<Double> axis;
```
c0d3rman: Now that this is a general command, this isn't necessarily an `axis` so we should refactor this to something else (maybe `supplier`).

commands/ThresholdCommand.java
```diff
@@ -0,0 +1,44 @@
...
+	protected final double threshold;
+	protected final boolean invert;
+
+	public ThresholdCommand(Command command, Supplier<Double> axis, double threshold, boolean invert) {
```
c0d3rman: Is it worth it to make this work with other comparable types, not just `Double`? (i.e. `Integer`) I'm not sure.

commands/ThresholdCommand.java
```diff
@@ -0,0 +1,44 @@
...
+		this.threshold = threshold;
+		this.invert = invert;
+	}
+
+	public ThresholdCommand(Command command, Supplier<Double> axis, double threshold) {
+		this(command, axis, threshold, false);
+	}
+
+	protected boolean pastThreshold(double value) {
+		return (!invert && value >= threshold) || (invert && value <= threshold);
```
c0d3rman: This is pedantic, but I think we can do this without doing the `value >= threshold` twice:
```
(value >= threshold) ^ invert
```
where the `^` is XOR (exclusive or, one or the other but not both).
We want to return true if the value is greater than the threshold or if we're inverted but not both.

commands/ThresholdCommand.java
```diff
@@ -0,0 +1,44 @@
...
+		this.axis = axis;
+		this.threshold = threshold;
+		this.invert = invert;
+	}
+
+	public ThresholdCommand(Command command, Supplier<Double> axis, double threshold) {
+		this(command, axis, threshold, false);
+	}
+
+	protected boolean pastThreshold(double value) {
```
c0d3rman: Should this take in a value or should it directly call `axis.get()`?

commands/ThresholdCommand.java
```diff
@@ -0,0 +1,44 @@
...
+		return (!invert && value >= threshold) || (invert && value <= threshold);
+	}
+
+	@Override
+	protected void execute() {
+		if (pastThreshold(axis.get()) && !command.isRunning()) {
```
c0d3rman: @carturn did we experience weirdness with `isRunning` in the past? I seem to recall some issue with it not being triggered the first tick?

commands/ThresholdCommand.java
```diff
@@ -0,0 +1,44 @@
...
+			command.start();
+		} else if (!pastThreshold(axis.get()) && command.isRunning()) {
+			command.cancel();
+		}
+	}
+
+	@Override
+	protected boolean isFinished() {
+		return false;
```
c0d3rman: Perhaps we should add a version that, when the supplier reaches the threshold, starts the command and ends immediately? (And one that does the same thing with cancelling a command?)

billwpierce:
@c0d3rman I fixed the problems.

commands/ThresholdCommand.java
```diff
@@ -0,0 +1,46 @@
...
+		this.threshold = threshold;
+		this.invert = invert;
+	}
+
+	public ThresholdCommand(Command command, Supplier<T> axis, T threshold) {
+		this(command, axis, threshold, false);
+	}
+
+	protected boolean pastThreshold() {
+		return ((supplier.get().compareTo(threshold) >= 0) ^ invert);
```
carturn: Do not use bitwise operators for logical operations. I recommend !=.

PR #190, opened 2017-03-14
Added a PID stop condition based on derivative

c0d3rman:
Added a stop condition to `CustomPIDController` based on how close the derivative is to zero. This is to prevent the loop ending as soon as it reaches zero error but is about to overshoot. It is also useful for P or PD loops, where the error doesn't converge to zero and the loop should end once the error is stable.


custom/motioncontrollers/CustomPIDController.java
```diff
@@ -319,4 +344,9 @@ public double get() {
 			return 0;
 		}
 	}
+
+	@Override
+	public boolean onTarget() {
+		return super.onTarget() && derivativeTolerance != 0 && Math.abs(lastErrorDerivative) < derivativeTolerance;
```
ajnadel: The derivative logic should be in its own function like onTarget()? I guess that's kind of what it is, but it seems kind of weird here.

billwpierce:
Is there a reason this hasn't been merged?

ajnadel:
@billwpierce as far as I'm concerned, this hasn't been well tested. We can do that and then merge?

oshlern:
We were using this on the 2018 bot and I don't know if it actually worked but it didn't break anything

PR #197, opened 2017-03-19
Made convenience functions for tuning a CustomPIDController on SmartDashboard

c0d3rman:


custom/motioncontrollers/CustomPIDController.java
```diff
@@ -350,4 +352,64 @@ public double get() {
...
+		// If we're missing any constants on SmartDashboard, put the current ones
+		if (!(SmartDashboard.containsKey(prefix + "_P") && SmartDashboard.containsKey(prefix + "_I")
+			&& SmartDashboard.containsKey(prefix + "_D") && SmartDashboard.containsKey(prefix + "_F"))) {
+			SmartDashboard.putNumber(prefix + "_P", getP());
+			SmartDashboard.putNumber(prefix + "_I", getI());
+			SmartDashboard.putNumber(prefix + "_D", getD());
+			SmartDashboard.putNumber(prefix + "_F", getD());
+			return;
+		}
+		setPIDF(SmartDashboard.getNumber(prefix + "_P", getP()),
```
ajnadel: Do you want to use the two-argument version with a default value? We've done this in the past while testing (with default 0).

carturn: I think this is a good way of doing it, it will default to the previous P/I/D/F value.

custom/motioncontrollers/CustomPIDController.java
```diff
@@ -350,4 +352,64 @@ public double get() {
...
+	/**
+	 * Put the PID controller to SmartDashboard for tuning.
+	 * Puts the error, setpoint, sensor value, and output.
+	 * This method adds a very small random value to everything
+	 * so that graphs show properly on SmartDashboard.
+	 * 
+	 * @param prefix
+	 *        The prefix to use when putting things on SmartDashboard.
+	 */
+	public void putToSmartDashboard(String prefix) {
```
carturn: This should be added as an abstract function to MotionController (same for the other functions).

ajnadel:
@carterturn Yes, perhaps, except that the values themselves are implementation-specific.

carturn:
@ajnadel Yes. I think we should only add abstract functions to MotionController. Otherwise we will have to declare all MotionControllers to actually be a certain non-abstract type.

custom/motioncontrollers/CustomPIDController.java
c0d3rman: That is what I'm doing...?

ajnadel: Ah yes -- I have... room to improve.

custom/motioncontrollers/CustomPIDController.java
```diff
@@ -350,4 +352,64 @@ public double get() {
...
+		double noise = (Math.random() - 0.5) * 0.0000001; // Generate very small noise centered at zero
+		SmartDashboard.putNumber(prefix + "_Error", getError() + noise);
+		SmartDashboard.putNumber(prefix + "_Setpoint", getSetpoint() + noise);
+		SmartDashboard.putNumber(prefix + "_Sensor", getSensorValue() + noise);
+		SmartDashboard.putNumber(prefix + "_Output", get() + noise);
```
ajnadel: Also, did we decide that the mutating effects of calling `get()` (of the I and D) calculations are not big enough to matter?

carturn: I think we have decided to fix that in the future. It should not make too much of a difference, as it does calculate time deltas every time.

ajnadel: Cool -- and the sensor updates fast enough that each few ticks will probably have a different value.

custom/motioncontrollers/CustomPIDController.java
```diff
@@ -350,4 +352,64 @@ public double get() {
...
+	 *        The prefix to use when putting things on SmartDashboard.
+	 */
+	public void updateFromSmartDashboard(String prefix) {
+		// If we're missing any constants on SmartDashboard, put the current ones
+		if (!(SmartDashboard.containsKey(prefix + "_P") && SmartDashboard.containsKey(prefix + "_I")
+			&& SmartDashboard.containsKey(prefix + "_D") && SmartDashboard.containsKey(prefix + "_F"))) {
+			SmartDashboard.putNumber(prefix + "_P", getP());
+			SmartDashboard.putNumber(prefix + "_I", getI());
+			SmartDashboard.putNumber(prefix + "_D", getD());
+			SmartDashboard.putNumber(prefix + "_F", getD());
```
ajnadel: D != F

ajnadel:
@carterturn better?

custom/motioncontrollers/BangBangController.java
```diff
@@ -18,6 +19,7 @@
 	protected double A;
 	protected double F;
 	protected double threshold;
+	protected static final String DEFAULT_SMARTDASHBOARD_PREFIX = "!!";
```
carturn: If this actually works, I think it is great. I would watch out for improper nonalphanumeric character parsing in NetworkTables though.

custom/motioncontrollers/CustomPIDController.java
c0d3rman: That is on the list to be fixed. (In fact I have a list of PID changes we need to make that various programmers will be doing before SVR)

custom/motioncontrollers/BangBangController.java
c0d3rman: Why did we change it to this? It's pretty jenk

carturn: This did not exist before. It is for the bang (!) bang (!) controller. We should make a shebang controller at some point.

c0d3rman: As funny as that is it should probably be `BangBang` or `BANGBANG`

c0d3rman:
I'm still not convinced we want to REQUIRE every controller to implement this. I think we should either have a default implementation or make this an interface.

carturn:
I think it is trivial enough to implement in most cases that a default implementation is not really going to make much of a difference. This is a convenient interface for tuning that should probably be preferred over the reupload with new constants method.

PR #203, opened 2017-04-21
More Motor Commands

carturn:
Adding some more motor commands after seeing some potential use cases.
Also trying to move to better constructors and double suppliers where possible (for more options in the future), thus adding the controller to double supplier adapter.

### DIFF
--- a/commands/motor/MotorControl.java
+++ b/commands/motor/MotorControl.java
@@ -23,6 +23,7 @@ public class MotorControl extends Command {
 	 * This Command directly controls a Motor based on an axis of the Controller.
 	 * This can allow an Operator to easily control a single Motor from an axis of the Controller.
 	 *
+	 * @param name
 	 * @param motor
 	 * @param controller
 	 * @param axis
@@ -31,8 +32,8 @@ public class MotorControl extends Command {
 	 *        A constant to add to the motor's speed
 	 *        Useful for using a controller for fine-tuning a constant speed
 	 */
-	public MotorControl(Motor motor, Controller controller, int axis, double scale, double offset) {
-		super("MotorControl");
+	public MotorControl(String name, Motor motor, Controller controller, int axis, double scale, double offset) {
+		super(name);
 		this.motor = motor;
 		this.controller = controller;
 		this.axis = axis;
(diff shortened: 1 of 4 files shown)


commands/motor/MotorControl.java
```diff
@@ -42,13 +46,39 @@ public MotorControl(Motor motor, Controller controller, int axis, double scale)
...
+	/**
+	 * This Command directly controls a Motor based on an axis of the Controller.
+	 * This can allow an Operator to easily control a single Motor from an axis of the Controller.
+	 *
+	 * @param motor
+	 * @param controller
+	 * @param axis
+	 * @param scale
+	 */
+	public MotorControl(Motor motor, Supplier<Double> speedSupplier, double scale) {
```
billwpierce: What is the point of this? Where does the scale get factored in/used?

commands/motor/MotorControl.java
```diff
@@ -14,25 +14,29 @@
...
 	 * @param axis
 	 * @param scale
+	 * @param offset
+	 *        A constant to add to the motor's speed
+	 *        Useful for using a controller for fine-tuning a constant speed
 	 */
-	public MotorControl(Motor motor, Controller controller, int axis, double scale) {
-		super("MotorControl");
+	public MotorControl(String name, Motor motor, Supplier<Double> speedSupplier, double scale, double offset) {
+		super(name);
```
billwpierce: This is pretty cool, however, we may run into compatibility problems with older versions. 10/10 would recommend for 2018-code, and maybe for 2017 if we take the time to do some refactoring.

commands/motor/MotorControl.java
billwpierce: What is the point of this? Where does the scale get factored in/used?

ajnadel: Yeah -- it looks like scale is just stored on the instance, and then never used. IMO scale is irrelevant when using speedSupplier; it was used when reading directly from the joystick, but with DoubleSupplier the user could just add the scale factor in the supplier itself.

Nice catch.

ajnadel:
This goes for the whole PR (and it's not critical), but there's a preexisting specialization of `Supplier<Double>`: [`DoubleSupplier`](https://docs.oracle.com/javase/8/docs/api/java/util/function/DoubleSupplier.html). Again, it's not critical, but we might as well use it.

commands/motor/MotorControl.java
HowardStark: Compatibility is something we rarely consider as we are not supporting a huge number of people on historical versions. If older software relies on something, a branch can be created to support it, or the older software can adapt.

billwpierce: Ok my B.

commands/motor/MotorSuppliedConstant.java
```diff
@@ -0,0 +1,89 @@
...
+	 * @param motor
+	 * @param speedSupply
+	 */
+	public MotorSuppliedConstant(Motor motor, Supplier<Double> speedSupply) {
+		this("MotorSuppliedConstant", motor, speedSupply, 1.0);
+	}
+
+	@Override
+	protected void initialize() {
+		speed = speedSupply.get() * scale + offset;
```
billwpierce: As @ajnadel mentioned, the scale is unnecessary if we can just adjust this in the supplier, no?

billwpierce:
> @ajnadel Can we remove scale from MotorControl?
@ajnadel I did, although before we merge I have two ideas/questions:

1. Should we use DoubleSupplier's instead? I'm happy to implement them, although I'd like to have a better understanding personally before we do.
2. Should we include the offset into the ControllerDoubleSupplier?

billwpierce:
Also, like someone should make sure I didn't mess anything up.

billwpierce:
I moved all of the scalings and offset into the supplier of the double, as that intuitively makes sense... we want the speed to be set in one location, not partially set in one location then finished being set and setting the motor to that speed in the other. That said if anyone knows of any places where this might cause compatibility issues, please double check before merging.

PR #204, opened 2017-09-17
BufferedOutputStream

leijurv:
this buffers the output stream. the buffer makes it so it doesn't do a file system write immediately each time, it builds them up and writes perhaps dozens of lines at a time.

### DIFF
--- a/LogKitten.java
+++ b/LogKitten.java
@@ -157,7 +157,6 @@ public class LogKitten {
 			try {
 				if (LogKitten.fileOutput != null) {
 					LogKitten.fileOutput.write(content.getBytes());
-					//LogKitten.fileOutput.flush();
 				} else {
 					System.out.println("Error logging: logfile not open");
 				}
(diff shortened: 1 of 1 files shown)


LogKitten.java
```diff
@@ -156,7 +157,7 @@ public static synchronized void logMessage(Object message, KittenLevel level, bo
 			try {
 				if (LogKitten.fileOutput != null) {
 					LogKitten.fileOutput.write(content.getBytes());
-					LogKitten.fileOutput.flush();
+					//LogKitten.fileOutput.flush();
```
HowardStark: If it's unneeded, it's unneeded. You can go ahead and delete the line.

LogKitten.java
```diff
@@ -46,7 +47,7 @@
 			// Create new file if it doesn't exist (this should happen)
 			file.createNewFile(); // creates if does not exist
 			// Create FileOutputStream to actually write to the file.
```
ajnadel: You might consider changing this comment but I think it's ok because the BufferedOutputStream wraps a FileOutputStream.

HowardStark:
This did not fix the competition issue of rapid logging making robot controls unresponsive, but it's a good start.

PR #214, opened 2018-01-20
Remove references to externally hosted files from travis

ajnadel:
No longer use t4904.xyz domain to host files required for building.

PR #215, opened 2018-02-06
fix bad CAN logic

ajnadel:
When we updated to the new CANJNI method signatures, some faulty logic was introduced. More specifically, `ByteBuffer`s filled using the #put() method were not rewound like they were when they came directly from CANJNI.


ajnadel:
Along with #214: Out of urgency, merging this without a secondary review.

PR #216, opened 2018-02-10
Logkitten improvements

quialst:
Added an instantiable class to LogKitten called Kitten. Kitten can in turn be extended and logged using LogKitten to be able to classify certain errors and log them in different files. Already created and integrated classes include CANKitten for logging CAN errors and messages and RobotKitten for logging based on which robot mode. Global logging is coded into the LogKitten class. MAKE SURE YOU REVIEW THE 2018 CODE PULL REQUEST AS WELL! This new code depends on it.

### DIFF
--- a/LogKitten.java
+++ b/LogKitten.java
@@ -13,7 +13,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
 import edu.wpi.first.wpilibj.DriverStation;
-import edu.wpi.first.wpilibj.hal.HAL;
+import edu.wpi.first.hal.HAL;
 import org.usfirst.frc4904.standard.Kitten;
 import org.usfirst.frc4904.standard.RobotModeKitten;
 


LogKitten.java
```diff
@@ -152,34 +212,76 @@ private static void reportErrorToDriverStation(String details, String errorMessa
...
 				ioe.printStackTrace();
 			}
 		}
+		for(Map.Entry<Kitten, String> i : kittenList.entrySet()) {
+			String content = getRobotMode() + " " + LogKitten.timestamp() + " " + level.getName() + ": " + LogKitten.getLoggerMethodCallerMethodName()
+			+ ": " + message + " \n";
+			switch(i.getValue()) {
+				case "RobotModeKitten":
+					if(getRobotMode() == ((RobotModeKitten)i.getKey()).getMode()) {
+						try {
```
HowardStark: There's a lot of duplicate code between "RobotModeKitten" and "CANKitten" cases. The code:
```
	try {
 		if(i.getKey().output != null) {
 			i.getKey().output.write(content.getBytes());
 		} else {
 			System.out.println("Error logging: " + i.getKey().category + " logfile not open");
		}
 	}
 	catch(IOException ioe) {
 		System.out.println("Error logging " + level.getName() + " message");
 		ioe.printStackTrace();
 	}
```

Could probably be reduced to a private method, or maybe a better flow could be worked out here.

LogKitten.java
```diff
@@ -26,44 +31,99 @@
...
+	}//static ends
+
+	/**
+	 * Gets the mode of the robot
+	 * 
+	 * @return a string with either auton or teleop
+	 */
+	private static String getRobotMode(){
+		String robotMode = null;
+		if(DriverStation.getInstance().isAutonomous()) {
```
HowardStark: Getting game-state information like this in a `static` method like this is dangerous. I don't know for certain that this won't work, so just make sure this gets tested on the bot before it gets merged.

LogKitten.java
HowardStark: Actually, it might be better instead to move the `logMessage` method to the `Kitten` class itself, and have `LogKitten` extend `Kitten` or something of that sort. Looking at a screen is making me nauseous right now but I can take a closer look tomorrow hopefully.

PR #217, opened 2018-02-11
MotorIdleGroup

NikhilSuresh24:
MotorIdle more than one motor. Useful for default command

### DIFF
--- a/commands/motor/MotorIdleGroup.java
+++ b/commands/motor/MotorIdleGroup.java
@@ -10,7 +10,7 @@ import edu.wpi.first.wpilibj.command.Subsystem;
  * Used as a default command
  *
  */
-public class IdleMotors extends CommandGroup{
+public class MotorIdleGroup extends CommandGroup{
 
 	/**
 	 * Run MotorIdle (from WPILib) in parallel on every motor
@@ -19,7 +19,7 @@ public class IdleMotors extends CommandGroup{
 	 * @param subsystem
 	 * @param motors
 	 */
-	public IdleMotors(Subsystem subsystem, Motor... motors) {
+	public MotorIdleGroup(Subsystem subsystem, Motor... motors) {
 		super("IdleMotors");
 		requires(subsystem);
 		for (Motor motor : motors) {
--- a/commands/motor/MotorIdleGroup.java
+++ b/commands/motor/MotorIdleGroup.java
@@ -1,8 +1,7 @@
 package org.usfirst.frc4904.standard.commands.motor;
 
 
-import org.usfirst.frc4904.standard.LogKitten;
-import org.usfirst.frc4904.standard.commands.KittenCommand;
+import org.usfirst.frc4904.standard.commands.motor.MotorIdle;
 import org.usfirst.frc4904.standard.subsystems.motor.Motor;
 import edu.wpi.first.wpilibj.command.CommandGroup;
 import edu.wpi.first.wpilibj.command.Subsystem;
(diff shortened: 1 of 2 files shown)


commands/motor/MotorIdleGroup.java
```diff
@@ -0,0 +1,58 @@
...
+
+	public static String makeName(Motor... motors) {
+		String name = "";
+		for (Motor motor : motors) {
+			if (motor.getName() != null) {
+				name += motor.getName() + " ";
+			} else {
+				name += "unnamed motor ";
+			}
+			name += "idling";
```
billwpierce: This "idling" is pointless.

commands/motor/MotorIdleGroup.java
```diff
@@ -0,0 +1,58 @@
...
+		this(MotorIdleGroup.makeName(motors), subsystem, motors);
+	}
+
+	public static String makeName(Motor... motors) {
+		String name = "";
+		for (Motor motor : motors) {
+			if (motor.getName() != null) {
+				name += motor.getName() + " ";
+			} else {
+				name += "unnamed motor ";
```
billwpierce: no need to put "unnamed motor", just put "motor"

commands/motor/MotorIdleGroup.java
```diff
@@ -0,0 +1,58 @@
...
+	 * Logs the names of all motors that are being idled
+	 *
+	 * @param subsystem
+	 * @param motors
+	 */
+	public MotorIdleGroup(Subsystem subsystem, Motor... motors) {
+		this(MotorIdleGroup.makeName(motors), subsystem, motors);
+	}
+
+	public static String makeName(Motor... motors) {
```
billwpierce: Maybe restrict the naming size, so only the first couple of motors are used in naming.

commands/motor/MotorIdleGroup.java
```diff
@@ -0,0 +1,58 @@
...
+	/**
+	 * Run MotorIdle (from WPILib) in parallel on every motor
+	 * from a list of motors with unrestricted size.
+	 * Logs the names of all motors that are being idled
+	 *
+	 * @param subsystem
+	 * @param motors
+	 */
+	public MotorIdleGroup(Subsystem subsystem, Motor... motors) {
+		this(MotorIdleGroup.makeName(motors), subsystem, motors);
```
billwpierce: Maybe we just want to do "unnamed motor idle" instead of doing a fancy naming thing?

commands/motor/MotorIdleGroup.java
```diff
@@ -0,0 +1,58 @@
...
+	 * @param name
+	 * @param subsystem
+	 * @param motors
+	 */
+	public MotorIdleGroup(String name, Subsystem subsystem, Motor... motors) {
+		super(name);
+		requires(subsystem);
+		for (Motor motor : motors) {
+			requires(motor);
+			addParallel(new MotorIdle(motor));
```
billwpierce: Have we tested this?

ajnadel:
Thanks @NikhilSuresh24 for making this happen. That being said, I'm going to close this because of the way `standard` currently works:

Every motor controller on the robot in instantiated as a `SpeedController`, and convention says that it will be wrapped in an instance of the `Motor` class. `Motor` can be constructed with any number of SpeedControllers which is useful for doubling up on motors (like the arm this year).

`Motor` as a class [extends `Subsystem`](https://github.com/RoboticsTeam4904/standard/blob/master/subsystems/motor/Motor.java#L17), which means that it has a default command, [which happens to be `MotorIdle`.](https://github.com/RoboticsTeam4904/standard/blob/master/subsystems/motor/Motor.java#L171) The effect of this is that any `Motor`, if not running a command that requires it like `MotorConstant` or `ChassisMove`, will return to `MotorIdle`. **The reason I'm closing this PR is that we should use the default command on each `Motor` rather than constructing a new one.** The only reason a programmer would need a `MotorIdleGroup` is if they were fighting with another motor command, in which case they should just cancel the other command and let the default `MotorIdle` take over.

Let me know if there are any questions -- I'm happy to reconsider, though I'm pretty confident in this reasoning.

PR #224, opened 2019-01-10
Wait for

bcho04:
This branch adds a command `WaitFor()` in standard commands, which allows for an `addSequential()` call to wait for a specified amount of time.

```
class WaitFor extends Command {
    public WaitFor(double duration) { ... }
    public isFinished() { ... }
```


bcho04:
I am deleting the wait-for branch from standard, as it is no longer needed.

PR #225, opened 2019-01-17
made changes to replace deprecated methods in standard

NikhilSuresh24:
Changed several files to remove calls of deprecated methods

### DIFF
--- a/custom/sensors/NavX.java
+++ b/custom/sensors/NavX.java
@@ -1,7 +1,6 @@
 package org.usfirst.frc4904.standard.custom.sensors;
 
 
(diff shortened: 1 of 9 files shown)


custom/sensors/NavX.java
```diff
@@ -1,7 +1,6 @@
 package org.usfirst.frc4904.standard.custom.sensors;
 
 
-import org.usfirst.frc4904.standard.LogKitten;
```
Daniel-E-B: rather than getting rid of this, maybe consider logging navx errors?

NikhilSuresh24: this should probably be a diff branch

PR #229, opened 2019-02-07
Spark

NikhilSuresh24:
Code for sparks


custom/sensors/SparkEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+
+public class SparkEncoder implements CustomEncoder {
+    protected final CANEncoder encoder;
+    protected PIDSourceType pidSource;
+    protected double distancePerPulse;
+    protected boolean reverseDirection;
+    protected double truePosition;
+    protected double trueVelocity;
+
+    public SparkEncoder(CANSparkMax spark, boolean reverseDirection, double distancePerPulse) {
```
oshlern: make constructor with default direction

custom/sensors/SparkEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+        encoder = new CANEncoder(spark);
+        this.reverseDirection = reverseDirection;
+        this.distancePerPulse = distancePerPulse;
+        setPIDSourceType(PIDSourceType.kDisplacement);
```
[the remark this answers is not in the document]
oshlern: If not, add option to set source type in constructor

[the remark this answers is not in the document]
oshlern: The encoder appears to have a getVelocity (http://www.revrobotics.com/content/sw/max/sw-docs/SPARK-MAX-Java-API-Offline.pdf, page 13), but I think it's fine to default to positional PID.

custom/sensors/SparkEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+        return reverseDirection;
+    }
+
+    @Override
+    public void setReverseDirection(boolean reverseDirection) {
+        this.reverseDirection = reverseDirection;
+    }
+
+    @Override
+    public void reset() {
```
oshlern: Reset should likely affect both types of readings, whether or not they are being used for PID.
Currently, `pidSource==kDisplacement` will change `pidGet` as well as `getDistance`, but if it affects `getDistance` which may be called outside of PID it should likely be consistent and reset velocity too

custom/sensors/SparkEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+        if (pidSource == PIDSourceType.kDisplacement) {
+            truePosition = encoder.getPosition();
+        } else {
```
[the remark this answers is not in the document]
oshlern: nvm

custom/sensors/SparkEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+            trueVelocity = encoder.getVelocity();
+        }
+    }
+
+    @Override
+    public double pidGetSafely() {
```
oshlern: We utilize safely methods so that we an catch exceptions. Would trying to read from a spark encoder ever throw an exception? Unless you're super confident that it won't, then we should throw and catch errors in the same manner as CANEncoder

custom/sensors/SparkEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+
+import com.revrobotics.CANEncoder;
+import com.revrobotics.CANSparkMax;
+import org.usfirst.frc4904.standard.Util;
+import edu.wpi.first.wpilibj.PIDSourceType;
+
+public class SparkEncoder implements CustomEncoder {
+    protected final CANEncoder encoder;
+    protected PIDSourceType pidSource;
+    protected double distancePerPulse;
```
oshlern: SparkEncoders return data in rotations (or rotations per minute) [page 13](http://www.revrobotics.com/content/sw/max/sw-docs/SPARK-MAX-Java-API-Offline.pdf). You should change from using pulses to rotations

custom/sensors/SparkEncoder.java
oshlern: Because the encoders return data in rotations, we may want to keep the data unscaled, so there should be a default constructor with distancePerPulse = 1.0

custom/sensors/SparkEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+    public double pidGet() {
+        if (pidSource == PIDSourceType.kDisplacement) {
+            return getDistance();
+        } else {
+            return getRate();
+        }
+    }
+
+    @Override
+    public int get() {
```
oshlern: encoder returns data as a double representing rotations, so it wouldn't really make sense for this method to exist. Have we ever used get? should we modify it to fit this type of sensor as well (in CustomEncoder) or delete it?

custom/sensors/SparkEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+        return !reverseDirection == (encoder.getVelocity() >= 0);
+    }
+
+    @Override
+    public boolean getStopped() {
+        return Util.isZero(getRate());
+    }
+
+    @Override
+    public double getRate() {
```
oshlern: Consider converting from rotations (or distance) per minute to rotations (or distance) per second, or make that a configurable option, since I assume we will usually want it in seconds (except maybe for flywheels?)

custom/sensors/SparkEncoder.java
NikhilSuresh24: yeah, it never should be used, but its in custom encoder

custom/sensors/SparkEncoder.java
```diff
@@ -2,85 +2,196 @@
...
     @Override
     public double pidGet() {
-        if (pidSource == PIDSourceType.kDisplacement) {
-            return getDistance();
-        } else {
-            return getRate();
+        try {
+            return pidGetSafely();
+        }
+        catch (InvalidSensorException e) {
```
oshlern: The `InvalidSensorException` is thrown by our implementation of CANSensor at https://github.com/RoboticsTeam4904/standard/blob/master/custom/sensors/CANSensor.java#L59.
I don't think `CANEncoder.getPosition()` should ever throw the exception, so we don't need to check for it. Unless reading from the encoder would throw a different type of error or should throw that exception, this should be removed.

custom/sensors/SparkEncoder.java
```diff
@@ -95,40 +206,58 @@ public void setReverseDirection(boolean reverseDirection) {
...
+                return getRate();
+            default:
+                return getDistance();
+        }
     }
 
     @Override
-    public int getSafely() {
-        return get();
+    public int getSafely() throws InvalidSensorException {
```
oshlern: Can you add comment saying this is here because only of the bad super class requirement?

custom/sensors/SparkLimitSwitch.java
```diff
@@ -0,0 +1,54 @@
...
+    public void disable() throws CANMessageUnavailableException {
+        CANError status = limitSwitch.enableLimitSwitch(false);
+        if (status != CANError.kOK) {
+            throw new CANMessageUnavailableException("Limit Switch Status Not Ok");
+        }
+    }
+
+    @Override
+    public boolean get() {
+        return !limitSwitch.get();
```
oshlern: Why flip? Is that not accounted for in polarity (in the constructor)?

PR #231, opened 2019-02-14
Using the Correct XboxController Class

NikhilSuresh24:
makes CustomXbox use the correct xboxController class


oshlern:
y no build? half of travis says we need xbox for importing in customxbox controller

PR #232, opened 2019-02-18
Add readSafely method to CustomCAN

aemino:
Adds a readSafely method to CustomCAN.

### DIFF
--- a/custom/CustomCAN.java
+++ b/custom/CustomCAN.java
@@ -91,17 +91,17 @@ public class CustomCAN {
 	 *
 	 * @return byte[] (8 long)
 	 */
-	public byte[] read() throws CANMessageUnavailableException {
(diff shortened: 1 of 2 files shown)


Codadillo:
In the rest of our codebase the doSomethingSafely functions are the ones that throw errors and the doSomething functions are the ones with try catches.

custom/CustomCAN.java
```diff
@@ -91,7 +91,22 @@ public void writeSafely(byte[] data) {
...
+
+	/**
+	 * Reads data, returning an empty Optional if there is no available
+	 * message.
+	 * 
+	 * @return Optional<byte[]>
+	 */
+	public Optional<byte[]> read() {
+		try {
+			return Optional.of(readBuffer());
```
NikhilSuresh24: can this be readSafely instead of readBuffer so that read uses readSafely

PR #236, opened 2019-12-07
Rewrite getEnum() rounding for accuracy

### LINKED ISSUE #120: Incorrect xboxcontroller rounding
https://github.com/RoboticsTeam4904/standard/blob/43cd2c3a5227bb2859a68089059d6af95c8a8223/custom/controllers/XboxController.java#L153
```
angle = Math.round(angle / 45) * 45; // May have rounding errors. Due to rounding errors.
```
angle / 45 is integer division since angle is an integer and 45 is too. `angle = (int) Math.round((float) angle / 45F) * 45` may have been intended.

bcho04:
Fixed #120 with suggested formula. The results for side-to-side comparisons between the old and new formulas can be found at https://repl.it/repls/SuburbanSameDehardwarization.

### DIFF
--- a/custom/controllers/XboxController.java
+++ b/custom/controllers/XboxController.java
@@ -150,7 +150,7 @@ public class XboxController extends Joystick {
 		public static DPAD getEnum(int angle) {
 			angle = Math.abs(angle);
 			angle %= 360;
-			angle = Math.round(angle / 45) * 45; // May have rounding errors. Due to rounding errors.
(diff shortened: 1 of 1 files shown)


Daniel-E-B:
We need to check on travis first and then we should be able to merge

leijurv:
*sheds tear* i'm so proud

just 2 years and 11 months later :^)

PR #240, opened 2020-01-05
Update chassis submodule to wpilibj2

BenCantCode:
Updated ChassisShift and SolenoidShifters to extend the new CommandBase.


aemino:
I'm going to close this, since apparently we want to transition to using the `Command`/`Subsystem` interfaces, rather than the abstract `*Base` classes.

PR #242, opened 2020-01-05
Removing Health Checks

KarenKGuo:


KarenKGuo:
Undid changes to CommandRobotBase and CommandSendableChooser.

PR #244, opened 2020-01-05
Motor update

Krakavius:
Updated the motor commands and subsystems in standard for the new WPILib. Created a new CustomSpeedController class as a replacement for the effectively deprecated one in WPILib.


commands/Idle.java
```diff
@@ -28,19 +32,23 @@ protected static String joinSubsystemNames(Subsystem[] subsystems) {
...
-	protected void end() {}
+	public void end(boolean interrupted) {}
 
 	@Override
-	protected void interrupted() {}
-}
+	public Set<Subsystem> getRequirements() {
+		Set<Subsystem> motors = new HashSet<Subsystem>();
+		// TODO Auto-generated method stub
+		return motors;
```
CaseyManning: Possibly should be condensed to just { return new HashSet<Subsystem>(); }

commands/motor/MotorControlAccelCap.java
```diff
@@ -83,15 +83,19 @@ protected void execute() {
...
 		return false;
 	}
 
 	@Override
-	protected void end() {}
-
-	@Override
-	protected void interrupted() {
+	public void end(boolean interrupted) {
 		LogKitten.d("MotorControl interrupted");
```
CaseyManning: Again, I believe this code will make it log "MotorControl interrupted" whenever, the command ends, regardless of whether or not it was interrupted, and should probably be put inside an if statement.

commands/motor/MotorIdle.java
```diff
@@ -18,37 +21,36 @@
...
-	protected boolean isFinished() {
+	public boolean isFinished() {
 		return false; // default command
 	}
 
 	@Override
-	protected void end() {
-		LogKitten.d("MotorIdle ended");
+	public void end(boolean interrupted) {
+		LogKitten.d("MotorIdle interrupted");
```
CaseyManning: Same comment as in the other end methods.

commands/motor/MotorIdle.java
```diff
@@ -18,37 +21,36 @@
...
 	}
-
 	@Override
-	protected void interrupted() {
-		LogKitten.d("MotorIdle interupted");
+	public Set<Subsystem> getRequirements() {
+		Set<Subsystem> motors = new HashSet<Subsystem>();
+		motors.add(motor);
+		// TODO Auto-generated method stub
```
CaseyManning: Generally good practice to removd the TODO Auto-generated method stub when you have added the desired functionality to the method.

commands/motor/MotorPositionConstant.java
```diff
@@ -61,39 +61,44 @@ protected void initialize() {
...
 		// Please just die in a hole and never come back
-		if (fallbackCommand != null && fallbackCommand.isRunning()) {
+		if (fallbackCommand != null && fallbackCommand.isScheduled()) {
 			fallbackCommand.cancel();
 		}
 	}
-
+	
 	@Override
-	protected void interrupted() {}
```
CaseyManning: Again, it seems to me that this change will alter what happens when the command is interrupted, and I'm not sure if that was intentional.

commands/motor/MotorPositionControl.java
```diff
@@ -74,21 +74,25 @@ protected void execute() {
...
 		return false;
 	}
 
 	@Override
-	protected void end() {}
-
-	@Override
-	protected void interrupted() {
+	public void end(boolean interrupted) {
 		LogKitten.d("MotorPositionControl interrupted");
```
CaseyManning: Same as before, we don't want it to log "interrupted" when the command wasn't actually interrupted and just ended normally.

commands/motor/MotorSet.java
```diff
@@ -44,24 +44,26 @@ public void set(double speed) {
...
 	@Override
-	protected void end() {
+	public void end(boolean interrupted) {
 		motor.set(0);
 		LogKitten.d("MotorSet ended (motor speed set to 0)");
 	}
 
 	@Override
-	protected void interrupted() {
-		LogKitten.d("MotorSet interupted (motor speed undefined)");
```
CaseyManning: before, it would log when the command was interrupted, and you just removed that. it should probably be added back inside the end() method as `if(interrupted) {LogKitten.d("Motorset INterrupted ... ")}`

commands/motor/speedmodifiers/SetEnableableModifier.java
```diff
@@ -2,29 +2,34 @@
 
 
 import org.usfirst.frc4904.standard.subsystems.motor.speedmodifiers.EnableableModifier;
-import edu.wpi.first.wpilibj.command.Command;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import edu.wpi.first.wpilibj2.command.CommandScheduler;
 
-public class SetEnableableModifier extends Command {
+public class SetEnableableModifier extends CommandBase {
```
CaseyManning: I don't think i have a good understanding of this, so probably just ignore me, but is there a reason that this should extend commandbase instead of implementing command?

commands/motor/speedmodifiers/SetEnableableModifier.java
```diff
@@ -2,29 +2,34 @@
...
 	public SetEnableableModifier(boolean enable, EnableableModifier... modifiers) {
 		this.modifiers = modifiers;
 		this.enable = enable;
 		for (EnableableModifier modifier : modifiers) {
 			requires(modifier);
 		}
 	}
+	protected void requires(EnableableModifier modifier){
+
+	}
```
CaseyManning: Ok this seems wrong. You are calling the requires method, expecting it to do something, but the method is just empty. Did you accidentally add in your own requires method instead of changing the code to use the getRequirements method instead?

subsystems/motor/ServoSubsystem.java
```diff
@@ -70,7 +70,7 @@ public ServoSubsystem(Servo... servos) {
 		this("ServoSubsystem", servos);
 	}
 
-	@Override
+
```
CaseyManning: This seems a little concerning but I'm not actually sure if its wrong. Is there a reason that this can no longer override anything, and are you sure there isn't an alternate method you are supposed to override instead?

Krakavius: This was sort-of a last resort fix, we looked for a while but couldn't find an actual solution (such as a method to replace it with). I knew this wasn't the best fix, but it was the only thing we could find to fix the error, other than completely removing it.

commands/motor/MotorPositionControl.java
```diff
@@ -74,21 +74,27 @@ protected void execute() {
...
+			LogKitten.d("MotorPositionControl interrupted");
+		}
+	}
 	@Override
-	protected void interrupted() {
-		LogKitten.d("MotorPositionControl interrupted");
+	public Set<Subsystem> getRequirements() {
+		Set<Subsystem> motors = new HashSet<Subsystem>();
+		motors.add(motor);
+		// TODO Auto-generated method stub
```
Daniel-E-B: delete line

commands/motor/MotorPositionSet.java
```diff
@@ -61,30 +61,35 @@ protected void initialize() {
...
 	@Override
-	protected void end() {}
+	public void end(boolean interrupted) {}
 
 	@Override
-	protected void interrupted() {}
+	public Set<Subsystem> getRequirements() {
+		Set<Subsystem> motors = new HashSet<Subsystem>();
+		motors.add(motor);
+		// TODO Auto-generated method stub
```
Daniel-E-B: delete this line

subsystems/motor/ServoSubsystem.java
Daniel-E-B: what we are going with is completely removing `initDefaultCommand()`s and adding javadoc comments to the classes that they are from saying what the default command should be set to.

subsystems/motor/speedmodifiers/EnableableModifier.java
```diff
@@ -37,8 +36,4 @@ public boolean isEnabled() {
 		return enabled;
 	}
 
-	@Override
-	protected void initDefaultCommand() {
-		setDefaultCommand(new Idle(this));
```
Daniel-E-B: add a javadoc saying what to set the default command to

subsystems/motor/VelocitySensorMotor.java
```diff
@@ -1,48 +1,49 @@
 package org.usfirst.frc4904.standard.subsystems.motor;
 
-
 import org.usfirst.frc4904.standard.LogKitten;
 import org.usfirst.frc4904.standard.custom.motioncontrollers.MotionController;
 import org.usfirst.frc4904.standard.subsystems.motor.speedmodifiers.IdentityModifier;
 import org.usfirst.frc4904.standard.subsystems.motor.speedmodifiers.SpeedModifier;
-import edu.wpi.first.wpilibj.SpeedController;
+import org.usfirst.frc4904.standard.custom.motioncontrollers.CustomSpeedController;
```
Daniel-E-B: change all instances of `CustomSpeedController` to `SpeedController`. Update naming

subsystems/motor/speedmodifiers/EnableableModifier.java
```diff
@@ -1,10 +1,9 @@
 package org.usfirst.frc4904.standard.subsystems.motor.speedmodifiers;
 
 
-import org.usfirst.frc4904.standard.commands.Idle;
-import edu.wpi.first.wpilibj.command.Subsystem;
+import edu.wpi.first.wpilibj2.command.Subsystem;
 
-public class EnableableModifier extends Subsystem implements SpeedModifier {
+public class EnableableModifier implements Subsystem, SpeedModifier{
```
Daniel-E-B: extend `SubsystemBase`, set default command,

PR #247, opened 2020-01-07
Cancel updates

Exr0n:
Fixed `standard/commands/Cancel.java`

### DIFF
--- a/commands/Cancel.java
+++ b/commands/Cancel.java
@@ -3,18 +3,19 @@ package org.usfirst.frc4904.standard.commands;
 import java.util.Set;
 
 import org.usfirst.frc4904.standard.LogKitten;
-import org.usfirst.frc4904.standard.commands.CustomCommand;
+import edu.wpi.first.wpilibj2.command.CommandBase;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.Subsystem;
 
 /**
  * This class cancels a command
  */
-public class Cancel extends CustomCommand {
+public class Cancel extends CommandBase {
 	protected final Command command;
 
 	public Cancel(Command command) {
-		super("Cancel " + command);
+		super();
+		setName("Cancel " + command.getName());
 		this.command = command;
 	}
 
@@ -27,8 +28,7 @@ public class Cancel extends CustomCommand {
 		return true;
 	}
 
-	public Set<Subsystem> getRequirements()
-	{
+	public Set<Subsystem> getRequirements() {
 		return Set.of();
 	}
 }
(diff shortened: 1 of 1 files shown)


Exr0n:
**Please do not review, as the code changes are not up to standard as defined by Daniel on slack:**

> Hey everyone, a few notes about standard updates:
> for requirements, please use the following format (its a little different than what we were doing before, but this is what Rohan and the WPILib devs do): 
```java
 @Override
  public Set<Subsystem> getRequirements() {
    return Set.of(...requirements...);
  }
```
> `...requirements...` is a Vararg, so just provide a command separated list of subsystems (or a single subsystem with no commas)
> for `initDefaultCommands`, please delete the entire method and add a javadoc to the class that the method was formerly a member of saying what command should be set as the default. Default commands are now set when the subsystem is instantiated as subsystems no longer need to have knowledge of their default commands. If you have questions about javadocs, please ask me or nikhil.
> Generally avoid wildcard imports (*) as they clutter the local namespace. Please only import what you use.
> `command.start()` is now essentially `command.schedule()`. `command.initialize()` is still `command.initialize()`.

Exr0n:
Fixed the issues, please review :)

commands/Cancel.java
```diff
@@ -1,35 +1,34 @@
...
 		return true;
 	}
 
-	@Override
-	protected void end() {}
-
-	@Override
-	protected void interrupted() {}
+	public Set<Subsystem> getRequirements()
+	{
```
aemino: formatter formatter formatter plz

commands/Cancel.java
```diff
@@ -1,35 +1,34 @@
...
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.Subsystem;
+
+/**
+ * This class cancels a command
+ */
+public class Cancel extends CommandBase {
 	protected final Command command;
 
 	public Cancel(Command command) {
```
NikhilSuresh24: take in a `CommandBase`

commands/Cancel.java
```diff
@@ -1,35 +1,34 @@
...
 	}
 
-	@Override
-	protected void end() {}
-
-	@Override
-	protected void interrupted() {}
+	public Set<Subsystem> getRequirements() {
+		return Set.of();
+	}
```
NikhilSuresh24: delete

commands/Cancel.java
```diff
@@ -1,35 +1,34 @@
...
 		command.cancel();
 	}
 
-	@Override
-	protected void execute() {}
-
-	@Override
-	protected boolean isFinished() {
+	public boolean isFinished() {
 		return true;
```
NikhilSuresh24: can you make it return true only if `!command.isScheduled`

PR #248, opened 2020-01-07
Idles update

Exr0n:
I deleted IdleVerbosely at Nikhil's suggestion because it seemed to implement completely unrelated and outdated functionality.


Exr0n:
**Please do not review, as the code changes are not up to standard as defined by Daniel on slack.**

> Hey everyone, a few notes about standard updates:
> for requirements, please use the following format (its a little different than what we were doing before, but this is what Rohan and the WPILib devs do): 
```java
 @Override
  public Set<Subsystem> getRequirements() {
    return Set.of(...requirements...);
  }
```
> `...requirements...` is a Vararg, so just provide a command separated list of subsystems (or a single subsystem with no commas)
> for `initDefaultCommands`, please delete the entire method and add a javadoc to the class that the method was formerly a member of saying what command should be set as the default. Default commands are now set when the subsystem is instantiated as subsystems no longer need to have knowledge of their default commands. If you have questions about javadocs, please ask me or nikhil.
> Generally avoid wildcard imports (*) as they clutter the local namespace. Please only import what you use.
> `command.start()` is now essentially `command.schedule()`. `command.initialize()` is still `command.initialize()`.

Exr0n:
Fixed, please check :)

commands/Idle.java
```diff
@@ -1,46 +1,62 @@
...
+		}
+	}
 
-	@Override
-	protected void interrupted() {}
+	protected void interrupted() {
+		if (verbose) {
+			LogKitten.v("Idle " + getName() + " interrupted.");
+		}
+	}
```
NikhilSuresh24: delete this cuz interrupted no longer exists

commands/Idle.java
```diff
@@ -1,46 +1,62 @@
...
+	public boolean isFinished() {
+		if (verbose) {
+			LogKitten.v("Idle " + getName() + " isFinished?");
+		}
 		return false;
 	}
 
-	@Override
-	protected void end() {}
+	protected void end() {
```
NikhilSuresh24: `end` takes in a `boolean interrupted`

NikhilSuresh24: also call the superclass's implementation of `end`

commands/Idle.java
```diff
@@ -1,46 +1,62 @@
...
-		setInterruptible(true);
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import edu.wpi.first.wpilibj2.command.Subsystem;
+
+/**
+ * Requires subsystems so no other code can interact with them.
+ */
+public class Idle extends CommandBase {
+	boolean verbose;
+	Set<Subsystem> requirements;
```
NikhilSuresh24: delete

commands/Idle.java
```diff
@@ -1,46 +1,62 @@
...
+
+	public Idle(String name, boolean verbose, Subsystem... subsystems) {
+		super();
+		setName(name);
+		this.verbose = verbose;
+		requirements = Set.of(subsystems);
```
NikhilSuresh24: replace with `addRequirements(subsystems)`

commands/Idle.java
bcho04: not sure how to fix

PR #259, opened 2020-01-12
Update docblocks and formatting from #235

bcho04:
Fix docblocks, formatting based on suggestions in #235.

### DIFF
--- a/Util.java
+++ b/Util.java
@@ -1,6 +1,5 @@
 package org.usfirst.frc4904.standard;
 
-
 import edu.wpi.first.hal.util.BoundaryException;
 
 /**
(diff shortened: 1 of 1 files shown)


Util.java
```diff
@@ -13,12 +13,24 @@
 	 *
 	 * @param var
 	 *        The floating point number you want to compare
+	 * @param epsilon
+	 *		  The maximum value of var such that it can be considered zero.
```
aemino: I'd probably opt for "magnitude" rather than "value" here. Drop the `.` at the end for consistency. Also, it seems that the indentation is a bit messed up in this file; running the formatter should (hopefully) fix it.

Util.java
```diff
@@ -13,12 +13,24 @@
...
-	 * 		Whether or not it is within Double.MIN_VALUE of zero
+	 * 		Whether or not var is within epsilon of zero
 	 */
 	public static boolean isZero(double var, double epsilon) {
 		return Math.abs(var) < epsilon;
 	}
+
+	/**
+	 * A function for determining if floating point numbers are effectively zero.
+	 * Floating point arithmetic tends to introduce very small errors.
```
aemino: The original documentation for this function is a little meh. We could improve both instances of this summary to something like "Returns true if `{@code var}` is effectively zero. This is useful for floating point numbers, which tend to incur small errors through arithmetic operations." This is just a suggestion, so go with however you would prefer to phrase it.

Util.java
```diff
@@ -13,12 +13,24 @@
...
+	 *
+	 * @param var
+	 *        The floating point number you want to compare
+	 * @return
+	 * 		Whether or not var is within Math.sqrt(Math.ulp(1.0)) of zero
```
aemino: We don't need to be so explicit about the semantics here. Better to just leave it at "Whether or not `{@code var}` is effectively zero, determined by a reasonable heuristic" or something.

Util.java
```diff
@@ -13,12 +13,24 @@
 	 *
 	 * @param var
 	 *        The floating point number you want to compare
+	 * @param epsilon
+	 *		  The maximum value of var such that it can be considered zero.
 	 * @return
-	 * 		Whether or not it is within Double.MIN_VALUE of zero
+	 * 		Whether or not var is within epsilon of zero
```
aemino: "Whether or not `{@code var}` is within `{@code epsilon}` of zero

Util.java
```diff
@@ -13,12 +13,24 @@
...
 	public static boolean isZero(double var, double epsilon) {
 		return Math.abs(var) < epsilon;
 	}
+
+	/**
+	 * A function for determining if floating point numbers are effectively zero.
+	 * Floating point arithmetic tends to introduce very small errors.
+	 *
+	 * @param var
+	 *        The floating point number you want to compare
```
aemino: To fix the old again, this would be better phrased as "The floating point number to compare" in both instances

Util.java
```diff
@@ -1,26 +1,36 @@
...
-	public static boolean isZero(double var) {
-		return isZero(var, Math.sqrt(Math.ulp(1.0)));
+
+	/**
+	 * Returns true if {@code var} is less than {@code epsilon}. This is useful for
+	 * floating point numbers, which tend to introduce small errors when operated
+	 * on.
+	 *
+	 * @param value The floating point number to be compared
+	 * @return Whether or not {@code var} is effectively zero, determined by the ULP
```
aemino: Again, specifying the semantics of the heuristic used is not necessary because it shouldn't be relevant to the end user in this case.

Util.java
```diff
@@ -1,26 +1,36 @@
...
 import edu.wpi.first.hal.util.BoundaryException;
 
 /**
  * Common utilities
  */
 public class Util {
 	/**
-	 * A function for determining if floating point numbers are effectively zero.
-	 * Floating point arithmetic tends to introduce very small errors.
+	 * Returns true if {@code var} is less than {@code epsilon}. This is useful for
```
aemino: `{@code var}`s need to be updated to `{@code value}`

Util.java
```diff
@@ -1,26 +1,36 @@
...
+	 * floating point numbers, which tend to introduce small errors when operated
```
aemino: "operated on" sounds a bit odd

PR #261, opened 2020-01-12
Update formatting in all standard files

bcho04:
Updated formatting via `Option-Shift-F` for all files in standard to standardize docblocks and line length, among other things.


Daniel-E-B:
I'm going to hold off on merging this one until were done with standard updates on `2020-update`, because they will all need to be redone and their content will have changed

PR #282, opened 2023-03-17
Dev requirements subsystem base

Exr0n:
So that we don't forget to require the smallest subsystem and end up with overlapping commands.


Exr0n:
Because subsystems should require their composed subsystems, but command groups should not require their composed commands' requirements

PR #284, opened 2024-01-18
2024 standard

aze12345:
Same as for 2024

### DIFF
--- a/custom/motorcontrollers/CANTalonFX.java
+++ b/custom/motorcontrollers/CANTalonFX.java
@@ -1,12 +1,12 @@
 package org.usfirst.frc4904.standard.custom.motorcontrollers;
 
-import com.ctre.phoenix.motorcontrol.InvertType;
-import com.ctre.phoenix.motorcontrol.NeutralMode;
-import com.ctre.phoenix.motorcontrol.can.WPI_TalonFX;
+import com.ctre.phoenix6.signals.ForwardLimitValue;
+import com.ctre.phoenix6.signals.NeutralModeValue;
+import com.ctre.phoenix6.signals.ReverseLimitValue;
+import com.ctre.phoenix6.hardware.TalonFX;
 
-public class CANTalonFX extends WPI_TalonFX implements TalonMotorController {
-	protected static final NeutralMode DEFAULT_NEUTRAL_MODE 	= NeutralMode.Coast;
-	protected static final InvertType  DEFAULT_INVERT_TYPE  	= InvertType.FollowMaster;
+public class CANTalonFX extends TalonFX {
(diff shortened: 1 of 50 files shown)


custom/sensors/CustomCANCoder.java
```diff
@@ -1,23 +1,25 @@
...
+import com.ctre.phoenix6.hardware.CANcoder;
+import com.ctre.phoenix6.configs.CANcoderConfiguration;
+// import com.ctre.phoenix.sensors.SensorTimeBase;
+// TODO: what is sensor time base why is it gone
 
 // import org.usfirst.frc4904.standard.LogKitten;
 import org.usfirst.frc4904.standard.Util;
 
-public class CustomCANCoder extends CANCoder implements CustomEncoder {
+public class CustomCANCoder extends CANcoder implements CustomEncoder {
```
Papayapaya, line 11: is CANcoder one word? the C is lowercase everywhere except the class name

custom/motorcontrollers/CANTalonFX.java
```diff
@@ -1,12 +1,12 @@
...
+import com.ctre.phoenix6.signals.ForwardLimitValue;
+import com.ctre.phoenix6.signals.NeutralModeValue;
+import com.ctre.phoenix6.signals.ReverseLimitValue;
+import com.ctre.phoenix6.hardware.TalonFX;
 
-public class CANTalonFX extends WPI_TalonFX implements TalonMotorController {
-	protected static final NeutralMode DEFAULT_NEUTRAL_MODE 	= NeutralMode.Coast;
-	protected static final InvertType  DEFAULT_INVERT_TYPE  	= InvertType.FollowMaster;
+public class CANTalonFX extends TalonFX {
+	protected static final NeutralModeValue DEFAULT_NEUTRAL_MODE 	= NeutralModeValue.Coast;
```
annvoor, line 9: should change that comment block, inverted is not a thing for the falcons here anymore i don't think

custom/sensors/CustomCANCoder.java
annvoor: for future reference, i think it's CANcoder and not CANCoder bc the ctre imports are CANcoder, but i don't want to change tho whole class name and stuff so it stays

custom/sensors/CustomCANCoder.java
```diff
@@ -82,7 +84,8 @@ public boolean getDirection() {
 
     @Override
     public boolean getDirectionSafely() throws InvalidSensorException {
-        return !reverseDirection == (getPosition() >= 0);
+        // getPosition() returns status signal, can't use >= on status signal
+        return !reverseDirection == (getPosition().getValue() >= 0);
```
annvoor, line 88: i think you can do getPosition().getValue() and then change what the function is returning to smth else and it works? i don't remember but it works in a different file (idk which one)

